PR #366, opened 2012-01-02
Initial support for lists of language strings

dumptruckman:
The default message providers still only returns one string, but it stores messages in a list.


dumptruckman:
This should probably just be redone in a different way at this point.

Issue #912, opened 2012-10-15
Lack of communication 


Azurago:
I feel as though this plugin team has lost touch with the community. Lack of communication and updates are becoming very frustrating. If this comment section isn't looked at by the dev, than disable them here and tell us where to go. If the plugin isn't being supported as heavily, tell us. I rely very heavily on this plugin and its sad to see these pages with no change every week I check.

Its unsettling to have to use dev builds on my bigger plugins. All I ask is a little more clarity on whats going on. Sorry to have gone on a little tangent, but I am losing faith in an otherwise great plugin.

PR #927, opened 2012-10-22
Replaced respawnworld and bedspawn with respawnDestination

main--:
Not tested at all
Experimental
Handle with care
I'm tired

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/configuration/SpawnLocation.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/configuration/SpawnLocation.java
@@ -82,6 +82,24 @@ public class SpawnLocation extends Location implements ConfigurationSerializable
         return serialized;
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final Location other = (Location) obj;
+
+        return /* (this.getWorld() == other.getWorld() || (this.getWorld() != null && this.getWorld().equals(other.getWorld())))
+                && */ (Double.doubleToLongBits(this.getX()) == Double.doubleToLongBits(other.getX()))
+                && (Double.doubleToLongBits(this.getY()) == Double.doubleToLongBits(other.getY()))
+                && (Double.doubleToLongBits(this.getZ()) == Double.doubleToLongBits(other.getZ()))
+                && (Float.floatToIntBits(this.getPitch()) == Float.floatToIntBits(other.getPitch()))
+                && (Float.floatToIntBits(this.getYaw()) == Float.floatToIntBits(other.getYaw()));
+    }
+
     /**
      * Let Bukkit be able to deserialize this.
      * @param args The map.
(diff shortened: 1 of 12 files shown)


src/main/java/com/onarandombox/MultiverseCore/configuration/SpawnLocation.java
```diff
@@ -82,6 +82,24 @@ public Block getBlock() {
...
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final Location other = (Location) obj;
+
+        return /* (this.getWorld() == other.getWorld() || (this.getWorld() != null && this.getWorld().equals(other.getWorld())))
```
FernFerret: Why is world equality not checked?

FernFerret: Nevermind! This is a location already derived in a world.

main--: First of all, it sure is brave to ask this question now that you went full valve time and defined "tonight" as 3.4 years :smile: . Still not beating valve's 4.7 years record though.

Surprise, surprise, long after esper.net has dropped both my original account and its reincarnation, I'm actually still alive and responding.

I guess you can understand that I literally have no memory of this place after all this time, so I can only guess what I was trying to do here. After reading the entire sourcefile, it looks like a `SpawnLocation`'s `worldRef` is more like a hint - it's just a weak reference after all. I did this for whatever reason - perhaps it was related to serialization? Or maybe to avoid redundancy because a "spawn location" is by definition always tied to a specific world? Anyways, since this looks a lot like a cache, checking equality probably doesn't really make sense. Then again, I committed commented-out code without a specific reason, so chances are I really didn't think this through.

---

Now to avoid getting your hopes up: I like this project a lot and spent a lot of time on it mainly because it was fun. In fact, most of my contributions happened after I had already stopped playing Minecraft. There was that time when I was fixing basically every issue but that was time-consuming and the day came when I couldn't keep it up. #912 came but didn't really take off. I considered resuming work on this a few times but never really got to it. Today, none of the machines I own even has Java installed and I'd like to keep it that way - not to mention that my time constraints really aren't improving.

tl;dr So Long, and Thanks for All the Fish! :heart: (And it was a lot more fish than you'd imagine - believe me.) I might pop in every once in a while and ramble about how I have no idea what I was thinking 4 years ago like I'm doing right now but I doubt that I'll be a useful help any time soon.

main--: >Nevermind! This is a location already derived in a world.

Ah, the joys of responding too slowly. Whatever. ¯\_(ツ)_/¯

FernFerret: No worries @main--!

I've ported almost all of this over to the new codebase (WorldProperties class) and am making progress. Found a few issues, but nothing insurmountable!

Thanks again for all of your work o/

benwoo1110:
outdated PR

PR #1021, opened 2012-12-26
Custom teleporters.

main--:
This would allow Multiverse/Multiverse-Portals#222. But I'm not sure if this is the best way to do it.


jAvEE-fr:
Why delete this? how can we make portals ot teleport throught Bungeecord?

nicegamer7:
I deleted the old branch because the stuff it was adding got accepted into MV. How is it used? I'm not sure...

jAvEE-fr:
Oh...
thank you for this response so fast... it will be so great to have mvtp or multiverse portal to send players to another server by bungeecord... Because it will be another plugin off my server... good for small servs...

PR #1135, opened 2013-03-28
Add API and implementation for adding and removing MV worlds without loading/unloading.

cybertiger:
Other plugins may wish to add MV worlds which they have loading using a custom world loader to the MV config, and later remove them.

They should be able to do this without causing Mutliverse to attempt to load them.


benwoo1110:
outdated PR

Issue #1096, opened 2013-02-21
Permissions * node always supersedes individual permissions?


the-sane:
I'm using PermissionsBukkit 2.0 on a 1.4.7 server.  With my other plugins, for example WorldGuard, I'm able to use the \* node followed by individual subsets of that node to create either a whitelist or a blacklist of permissions.  When I try to do this in multiverse with the enforce access feature enabled, the individual permission seems to be ignored, as if it's completely superseded by the \* permission above it.

For example, I am trying to do this:

multiverse.portal.access.*: true
multiverse.portal.access.restrictonlythisone: false

I'm posting in the MVCore area because the same thing happens when enforcing access to worlds, not just portals.  As I said above, I've been able to do this with other plugins successfully but I'm not sure if that means they are implementing that part of the permissions API incorrectly, or if multiverse is.

I prefer the whitelist/blacklist approach because it does not require listing out every single item in the permissions list (in this case, every portal and world), and it also doesn't require an administrator to update the permissions whenever a new world or portal is added.  It would only need to be updated if the new world/portal in question needs to be restricted, as in the example above.

Any help would be greatly appreciated.

PR #1158, opened 2013-04-13
Broken perms?

main--:
As pointed out in #1096, parent-child permission relationships are a bit weird: The parent permission **overrides** the child permission!

The work in this branch is not finished yet, so I don't want this to be pulled now but I'm looking for some feedback on whether this is a good or a bad approach to solve the issue.


benwoo1110:
outdated PR

PR #1556, opened 2015-03-04
Update Maven POM

MattBDev:
Updates all versions of plugins and dependencies.
Stop including Bukkit because it really isn't needed and annoying for people using multiverse as a dependency.
Stops including Vault and only it's API because using Vault directly is bad.
Scopes JUnit to test

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,8 +17,8 @@
             <url>http://repo.onarandombox.com/content/groups/public</url>
         </repository>
         <repository>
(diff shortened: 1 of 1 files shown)


tuskiomi:
is their maven broken?

Rigby90:
Was closing/reopening to trigger CircleCI and TravisCI for an updated build output.

I can look over this tonight.

tuskiomi:
Thanks, Rigby.

PR #1581, opened 2015-06-10
Update pom.xml

Rigby90:
Pull request for CircleCI to build and test :(

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,10 +26,10 @@
             <url>http://repo.mcstats.org/content/repositories/public</url>
         </repository>
         <!-- Required for Vault -->
-        <!--repository>
-            <id>herocraft</id>
-            <url>http://ci.herocraftonline.com/plugin/repository/everything/</url>
-        </repository-->
+        <repository>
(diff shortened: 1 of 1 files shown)


dumptruckman:
Problem appears to be getting craftbukkit, not vault.

Rigby90:
Indeed, back at home now so was going to take a proper look.

Rigby90:
@dumptruckman Ok, looks like it struggles to get CraftBukkit because of your Logging library.

https://github.com/dumptruckman/Logging/blob/master/pom.xml

I'm able to add the Essentials repo to grab the version required for building. Any reason why that relies on CraftBukkit and not Bukkit? Only saw a reference to the Plugin class.

dumptruckman:
Huh. I forgot that was a separate lib. Anyway, I updated it to use Bukkit instead. Build successful on our ci.

Rigby90:
Nice, unable to get the tests to pass on my PC. Not sure why, been too long!

TWSSYesterday:
Fixed in #1586 ?

PR #1586, opened 2015-06-29
Update pom.xml

TWSSYesterday:
Should fix compilation issues. Worked for me.


dumptruckman:
This may make it work but it is not a good solution as this disables our  javadocs

TWSSYesterday:
True

PR #1618, opened 2015-10-23
Fix Travis-CI failing for Codehaus Dependencies

Rigby90:


Rigby90:
@dumptruckman I'm sure there's a better way but until then...

FernFerret:
Woah! There's a name I haven't seen in a while! :wave: @Rigby90

PR #1652, opened 2016-01-30
Added Spigot Dependency and repository

ThunderGemios10:


TWSSYesterday:
Looks good, a lot of people won't need the higher version though. Possibly leaving it for a short while. That being said, the Spigot repository is a lot better to use as it's most certainly one of the more stable of repositories out there.

ThunderGemios10:
@TWSSYesterday It is. You should probably change it soon. as well for other Multiverse repo

PR #1664, opened 2016-03-02
Updated to the latest Spigot 1.9 version.

ThunderGemios10:


ThunderGemios10:
@dumptruckman @TWSSYesterday This is really required to be merged as this fixes tons of bugs at Spigot 1.9

dumptruckman:
This is unnecessary and not correct.

ThunderGemios10:
@dumptruckman How? Can you please clarify? 1.9 R0.1 is different.

Unless they are the same build?

dumptruckman:
1.9 R0.1 did not change anything that matters to us so it makes absolutely no difference in our build process. The only reason I updated it to 1.9 at all is because 1.9 removed a particular deprecated method signature and I want to build against 1.9 to be sure we are compatible with the updated method signature.

ThunderGemios10:
@dumptruckman I see, Thank you for your explanation.

PR #1670, opened 2016-03-06
Spigot-1.9

nichlasm1:
I really need MV-Core and the other addon plugins updated to 1.9.

Since the source code can't be imported i can't do this by myself so i would like to request this.


FernFerret:
Wait what... What is this even... @dumptruckman has already updated core for 1.9... In fact, I'm running it myself, please see our build page here for downloads: http://ci.onarandombox.com/view/Multiverse/job/Multiverse-Core/716/

Not merging, because I don't know what this is, but it is a hot mess.

PR #1746, opened 2016-08-03
Fix failing tests

rlf:


FernFerret:
Thanks for this! I'll be adding this soon along with @httpdop's changes too. It won't be the exact hash of this, so this PR will probably get closed, but rest assured you'll get credit :)

I can't believe a HashMap was used, reversed and then thought iteration would be "ok". I bet we were just getting lucky. I've run the `git blame` but I'll leave that for the code gremlins to discover :)

Here: have this: 🍺 

PS: You made the files executable.

FernFerret:
Closing this PR since I've put the changes on master. Thanks again.

rlf:
> PS: You made the files executable.

Haha, I'm using Git Bash on Windoze, not sure how/why it changed the filemode, wasn't intentional.
No need for credit, just like to get stuff working :)

PR #1794, opened 2017-02-01
RFC:  Teleport to players' last locations in worlds

qwylz:
First, thank you for the great plugin.  Recently I tested several plugins for managing multiple worlds and Multiverse-Core handily came out on top.  Thank you!

Secondly, this is not actually a request to merge code at this time--this code is definitely not ready to be merged.  I've submitted this PR more as an early request for comments and guidance as I work to get the code ready for a real pull request.  Will you please read through this submission and the accompanying code and provide any suggestions and guidance necessary for me to get this ready for a true pull request?

This code is intended to add a new, non-default feature to Multiverse-Core.  When enabled, the feature makes it so that when a player leaves a world, his/her last position in the world is saved.  Then, when teleporting back to the world using the teleport syntax of just the bare world name (i.e. no "w:" or "e:" prefix), the player is teleported to his/her previously saved location in the world rather than to the world's spawn point.  If the player has no saved location for the world, or if the save cannot be read/parsed for some reason, the player is teleported to the world's spawn point as a fallback measure.  The behavior when using the "w:" or "e:" syntaxes is unchanged (i.e. always teleports to the world's spawn location or an exact location, respectively).

In it's current form, the code works, but is in an hackish, "can I even get this to work" kind of state.  I need to take care of the following items before it's really ready for a true pull request:

1)  Currently, it's not possible to disable the feature.  I still need to add a configuration property for the plugin that will control whether the feature is enabled or disabled.  I plan to have it default to being disabled, so that, by default, there is no change in behavior.

2)  The code is currently embedded in several different classes.  Very ugly.  I need to refactor the code into it's own class(es) (and an exception) which will be utilized by the various listener and command classes.

3)  I haven't done anything about permissions, yet.  There should probably be separate permissions for teleporting to a world at the player's saved location versus teleporting to the world's spawn point or exact locations.

4)  The code currently saves a players location in a world when leaving a world either via teleporting or the player quitting.  It does not save players' locations when the server is stopped.  I'm not sure how to listen for that.  Is there a bukkit event?  A player's location is also not saved when he/she is kicked from a world.  I figure that if someone is misbehaving enough to be kicked, his/her location doesn't need to be save (however, if there was a previously saved location for the player, that saved information is not deleted).

5)  While my day job is software development, my experience is in other languages.  I've never written a line of Java until writing the code for this feature yesterday.  Hopefully, I haven't abused Java too badly, and there's likely more Java-like ways to do some of what this code does.  Suggestions are welcome, please.

One final note:  I have developed this code under Oracle's official Java 8 JDK.  Hopefully, I haven't used any constructs or patterns that cause problems on earlier versions or other flavors of Java.

Again, thank you for the great plugin.  I hope that I can get the code for this feature in order so that it can be merged.  Please let me know of any guidance and suggestions that you have.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
@@ -59,6 +59,7 @@ import com.onarandombox.MultiverseCore.destination.BedDestination;
 import com.onarandombox.MultiverseCore.destination.CannonDestination;
 import com.onarandombox.MultiverseCore.destination.DestinationFactory;
 import com.onarandombox.MultiverseCore.destination.ExactDestination;
+import com.onarandombox.MultiverseCore.destination.LastLocationDestination;
 import com.onarandombox.MultiverseCore.destination.PlayerDestination;
 import com.onarandombox.MultiverseCore.destination.WorldDestination;
 import com.onarandombox.MultiverseCore.event.MVVersionEvent;
@@ -459,6 +460,7 @@ public class MultiverseCore extends JavaPlugin implements MVPlugin, Core {
         this.destFactory.registerDestinationType(WorldDestination.class, "");
         this.destFactory.registerDestinationType(WorldDestination.class, "w");
         this.destFactory.registerDestinationType(ExactDestination.class, "e");
+        this.destFactory.registerDestinationType(LastLocationDestination.class, LastLocationDestination.getID());
         this.destFactory.registerDestinationType(PlayerDestination.class, "pl");
         this.destFactory.registerDestinationType(CannonDestination.class, "ca");
         this.destFactory.registerDestinationType(BedDestination.class, "b");
--- a/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
@@ -208,6 +208,7 @@ public class MultiverseCore extends JavaPlugin implements MVPlugin, Core {
     private CommandHandler commandHandler;
 
     private static final String LOG_TAG = "[Multiverse-Core]";
+    public  static final String PLAYER_LOCATION_DATA = "player_location_data";
 
     // Multiverse Permissions Handler
     private MVPermissions ph;
--- a/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
@@ -76,6 +76,7 @@ import com.onarandombox.MultiverseCore.utils.AnchorManager;
 import com.onarandombox.MultiverseCore.utils.MVEconomist;
 import com.onarandombox.MultiverseCore.utils.MVMessaging;
 import com.onarandombox.MultiverseCore.utils.MVPermissions;
+import com.onarandombox.MultiverseCore.utils.MVPlayerLocation;
 import com.onarandombox.MultiverseCore.utils.MVPlayerSession;
 import com.onarandombox.MultiverseCore.utils.SimpleBlockSafety;
 import com.onarandombox.MultiverseCore.utils.SimpleLocationManipulation;
@@ -208,7 +209,6 @@ public class MultiverseCore extends JavaPlugin implements MVPlugin, Core {
     private CommandHandler commandHandler;
 
     private static final String LOG_TAG = "[Multiverse-Core]";
-    public  static final String PLAYER_LOCATION_DATA = "player_location_data";
 
     // Multiverse Permissions Handler
     private MVPermissions ph;
@@ -259,6 +259,8 @@ public class MultiverseCore extends JavaPlugin implements MVPlugin, Core {
         // Setup our SafeTTeleporter
         this.safeTTeleporter = new SimpleSafeTTeleporter(this);
         this.unsafeCallWrapper = new UnsafeCallWrapper(this);
+        // Set up MVPlayerLocation
+        MVPlayerLocation.init(this);
     }
 
 
--- a/src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
@@ -10,20 +10,16 @@ package com.onarandombox.MultiverseCore.destination;
 import com.onarandombox.MultiverseCore.MultiverseCore;
 import com.onarandombox.MultiverseCore.api.MVDestination;
 import com.onarandombox.MultiverseCore.commands.TeleportCommand;
+import com.onarandombox.MultiverseCore.utils.MVPlayerLocation;
 import com.onarandombox.MultiverseCore.utils.PermissionTools;
 import com.pneumaticraft.commandhandler.Command;
-import org.bukkit.configuration.file.YamlConfiguration;
 import org.bukkit.entity.Player;
-import org.bukkit.Location;
 import org.bukkit.permissions.Permission;
 import org.bukkit.permissions.PermissionDefault;
 
-import java.io.File;
 import java.util.HashMap;
 import java.util.List;
-import java.util.logging.Level;
 import java.util.Map;
-import java.util.zip.DataFormatException;
 
 /** A factory class that will create destinations from specific strings. */
 public class DestinationFactory {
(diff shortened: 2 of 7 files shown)


dumptruckman:
1 & 3) Rather than a global config option, perhaps consider just using a single simple `default: false` permission for whether they should teleport to the world last location. This way the option can be per player and is still off by default.

2) I'll have to look over the code and make comments there.

4) PlayerQuitEvent should cover all cases of the player leaving the server. PlayerChangeWorldEvent should cover all cases of players changing world. I would suggest making a simple yaml data file for this information. Just be sure to associate the location data with player UUID.

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
```diff
@@ -37,19 +43,103 @@ public DestinationFactory(MultiverseCore plugin) {
...
+                } catch (Exception e) {
+                    this.plugin.log(Level.SEVERE, "Failed to load saved location of player '"
+                        + player.getName() + "' (" + playerID + ") in world '" + destination
+                        + "': " + e.getMessage() + ".");
+                    yc = null;
+                    // fall through to use world spawn location
+                }
+                if (yc != null) {
+                    try {
+                        if (! yc.isSet("schema"))
```
dumptruckman: YamlConfiguration should have like getInteger() for this.

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
```diff
@@ -37,19 +43,103 @@ public DestinationFactory(MultiverseCore plugin) {
...
+                try {
+                    yc.load(f);
+                } catch (Exception e) {
+                    this.plugin.log(Level.SEVERE, "Failed to load saved location of player '"
+                        + player.getName() + "' (" + playerID + ") in world '" + destination
+                        + "': " + e.getMessage() + ".");
+                    yc = null;
+                    // fall through to use world spawn location
+                }
+                if (yc != null) {
```
dumptruckman: It's been a while since I've looked at MV's code but I think we have a simpler way to serialize/deserialize location data.

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
```diff
@@ -37,19 +43,103 @@ public DestinationFactory(MultiverseCore plugin) {
...
+     * Gets a new destination from a string and possibly for a player.
      * Returns a new InvalidDestination if the string could not be parsed.
      *
      * @param destination The destination in string format.
+     * @param player The player.
      *
      * @return A non-null MVDestination
      */
-    public MVDestination getDestination(String destination) {
+    public MVDestination getDestination(String destination, Player player) {
```
dumptruckman: Probably better you leave the old method in just add this as a new method. This way we can preserve backwards compatibility.

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
qwylz: Ah ha!  Yes, I see that now.  When I first looked at the docs for YamlConfiguration, I didn't realize that the list of methods was not in alphabetic order.  I stopped when I saw the .get and .isSet methods and no other .get or .is methods.  I've refactored to use the .isInt, .isDouble, .getInt, and .getDouble methods.  Thanks for the tip!

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
qwylz: If you look a little later in the diff, you'll see that's what I already did.  I didn't want to break the existing API.  The original method is just a shim that supports the original API.  It simply calls the new API with null for player and the behavior is the same as the old behavior and still keeps the code DRY.

qwylz:
Having the ability to enable/disable this on a per-player basis is a good idea.  It allows for more use cases than my simple case which I had not thought of.  I think that using the permissions system might be a good way to handle how to enable/disable for each player.  I think that's what you were suggesting, as well (yes?).

I still think there will need to be a global default enable/disable item, otherwise I don't think my use case is achievable (unless I misunderstand what you're meaning).  My use case is to specifically allow this behavior for all users without having to enable it for each new player that comes along.  Having the global setting and then per-user permissions allows for both whitelisting and blacklisting use cases.  Out of the box, it would be in whitelisting mode so as to retain backward compatibility.

dumptruckman:
If you use only permissions (white list as you say) rather than also having a configuration value you can just give your default user the permission in your permission manager. Even better, if you want it for everyone no questions you can set it to `default: true` in permissions.yml in your server root.

qwylz:
I think my confusion comes from not having studied MV's permissions model, thus far.  Now that I've started to look at it, I *think* it's starting to make sense, but I need to do so more exploration on how the permissions all work together.

To that end, I have a question:

If a player has `multiverse.teleport.self.e`, it enables them to use `/mv tp e:` in/into any and all worlds for which they have `multiverse.access.WORLDNAME`, right?  In other words, if a player has the `.e` permission they can teleport to exact destinations in all worlds to which they have the corresponding `.access` permission and there is no way to say that a player can exact teleport in/to worlds A and C, but not B (but still allow the player access to B via other teleport methods), right?

dumptruckman:
Yes, you have that right.

qwylz:
Now that I understand how the permissions all work, I understand what you meant about using permissions rather than needing a global configuration item.  It's a great idea!  I've got that all sorted out and implemented now.  I think I've now taken care of all the outstanding items.

There are three new permissions used:

`multiverse.teleport.self.l`
`multiverse.teleport.other.l`
`multiverse.save.lastlocations`

The first two are used to control who can use the new last location destination functionality.  The `/mv tp` command, when given a destination that does not specify a location type identifier, will intelligently figure out whether to use a last location destination or just the original world spawn destination based on permissions.

If a player has either `multiverse.teleport.self.l` or `multiverse.save.lastlocations` his/her last location in a world will be saved (otherwise it will not be saved).  This makes it possible to make it so a user's last locations are saved without allowing them to teleport to their last locations (by just giving the user `multiverse.save.lastlocations`).

I think I have this all ready to be a real PR.  Can this code be merged?

P.S.
Sorry, I see that I spelled refactor wrong in my last commit.  :-(

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -258,6 +260,8 @@ public void onLoad() {
         // Setup our SafeTTeleporter
         this.safeTTeleporter = new SimpleSafeTTeleporter(this);
         this.unsafeCallWrapper = new UnsafeCallWrapper(this);
+        // Set up MVPlayerLocation
+        MVPlayerLocation.init(this);
```
dumptruckman: Is there any way we can avoid static initialization like this?

src/main/java/com/onarandombox/MultiverseCore/destination/LastLocationDestination.java
```diff
@@ -0,0 +1,228 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Location getLocation(Entity e) {
+        if (!this.isValid() || e == null || !(e instanceof Player)) {
+            return null;
+        }
+
+        player = (Player) e;
+        this.location = MVPlayerLocation.getPlayerLastLocation(player, world);
```
dumptruckman: Can this location be set in setDestination using a non static method? Such as `plugin.getPlayerLocationManager().getPlayerLastLocation...` or something. Just trying to avoid static stuff like this.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -147,7 +149,13 @@ public void playerChangedWorld(PlayerChangedWorldEvent event) {
      */
     @EventHandler
     public void playerQuit(PlayerQuitEvent event) {
-        this.plugin.removePlayerSession(event.getPlayer());
+        Player player = event.getPlayer();
+        if (this.plugin.getMVPerms().hasPermission(player, "multiverse.teleport.self."
+                + LastLocationDestination.getID(), true)
```
dumptruckman: Theoretically the third arg here should be false, but it isn't used anymore so I guess it doesn't matter.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -234,6 +242,22 @@ public void playerTeleport(PlayerTeleportEvent event) {
         this.stateSuccess(teleportee.getName(), toWorld.getAlias());
     }
 
+    /**
+     * This method is called when a player teleports anywhere.
+     * @param event The Event that was fired.
+     */
+    @EventHandler(priority = EventPriority.MONITOR)
```
dumptruckman: You should add an `ignoreCancelled = true` here.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -315,7 +339,9 @@ private void sendPlayerToDefaultWorld(final Player player) {
             new Runnable() {
                 @Override
                 public void run() {
-                    player.teleport(plugin.getMVWorldManager().getFirstSpawnWorld().getSpawnLocation());
+                    LastLocationDestination destination = new LastLocationDestination();
```
dumptruckman: Why are there no permission checks here?

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+import com.onarandombox.MultiverseCore.api.MVDestination;
+import com.onarandombox.MultiverseCore.destination.ExactDestination;
+import org.bukkit.Location;
+import org.bukkit.configuration.file.YamlConfiguration;
+import org.bukkit.entity.Player;
+
+import java.io.File;
+import java.util.logging.Level;
+
+public class MVPlayerLocation {
```
dumptruckman: This name seems to suggest it belongs to a single player.

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+                + "' (" + playerID + ") was in world '" + world + "' at "
+                + Double.toString(location.getX()) + ", "
+                + Double.toString(location.getY()) + ", "
+                + Double.toString(location.getZ()) + ", "
+                + Double.toString(location.getYaw()) + ", "
+                + Double.toString(location.getPitch()) + " before " + action + ".");
+
+        YamlConfiguration yc = new YamlConfiguration();
+        yc.set("schema", 1);
+        yc.set("world", world);
```
dumptruckman: Don't we have something that handles serializing and deserializing locations for configs?

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+        yc.set("y", location.getY());
+        yc.set("z", location.getZ());
+        yc.set("yaw", location.getYaw());
+        yc.set("pitch", location.getPitch());
+
+        File dir = new File(plugin.getDataFolder(),
+            PLAYER_LOCATION_DATA + File.separator + world);
+
+        try {
+            dir.mkdirs();
```
dumptruckman: The only exception this throws, `SecurityException`, is almost never going to be thrown. Instead you should be caring about the return value of this method.

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+        } catch (Exception e) {
+            plugin.log(Level.SEVERE, "Failed to create directory '"
+                + dir.toString() + "': " + e.getMessage());
+        }
+
+        try {
+            yc.save(new File(dir, playerID + ".yaml"));
```
dumptruckman: Extension should be `.yml`. Again, I think you could cut down on files by just having 1 file per player.

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+        String playerID;
+        File   file;
+
+        if (plugin == null || player.getUniqueId() == null)
+            return null;
+
+        playerID = player.getUniqueId().toString();
+
+        file = new File(plugin.getDataFolder(), PLAYER_LOCATION_DATA
+            + File.separator + world.getName() + File.separator + playerID + ".yaml");
```
dumptruckman: `.yml`

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private static class InvalidData extends Exception {
+        public InvalidData(String msg) {
+            super(msg);
+        }
+    }
+
+    public static void savePlayerLocation(Player player, Location location, String action) {
+        String world    = location.getWorld().getName();
```
dumptruckman: You should use World UUID rather than name. Also, I might suggest just having 1 file for each player (again using UUID) with all world data stored in it.

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private static final String PLAYER_LOCATION_DATA = "player_location_data";
+
+    private static MultiverseCore plugin = null;
+
+    public static void init(MultiverseCore p) {
+        plugin = p;
+    }
+
+    private static class InvalidData extends Exception {
```
dumptruckman: Exceptions should be follow the naming scheme SuchAndSuchException. Also, make sure there's not an existing exception in Java that might cover this case well enough.

src/main/java/com/onarandombox/MultiverseCore/utils/MVPlayerLocation.java
```diff
@@ -0,0 +1,159 @@
...
+            } catch (Exception e) {
+                plugin.log(Level.SEVERE, "Failed to load saved location of player '"
+                    + player.getName() + "' (" + playerID + ") in world '" + world.getName()
+                    + "': " + e.getMessage() + ".");
+                return null;
+            }
+            try {
+                if (! yc.isSet("schema"))
+                    throw new InvalidData("missing schema node");
+                if (! yc.isInt("schema"))
```
dumptruckman: `isInt` being true implies `isSet` is true. No need for both.

dumptruckman:
Also, consider squashing your commits.

Tzzzt123:
I vote for this Pull request!

ili-codes:
@Tzzzt123 Same here. Been trying to find a fix here on GH and see this. Would love for this to be a feature. Would help a lot.

FernFerret:
No need to really "vote" for this PR. @dumptruckman performed a review but
@qwylz has not yet responded.

tldr: It will be merged but some housekeeping needs to happen first.

mattysweeps:
3 years later

M4cs:
So how exactly do we enable this then?

mattysweeps:
https://github.com/Taronyuu/StayPut is almost what you'd want, except it doesn't choose the default spawn location when there's no previous location in the database.

edcous:
Any update on this?

LoneDev6:
Any news?

edcous:
Stay put

pixelblob:
Anything happening?

LoneDev6:
> 
> 
> Anything happening?

I don't think so

phemmer:
Stumbled across this PR while looking for how to preserve last location.
The functionality exists in the [Multiverse-Inventories plugin](https://dev.bukkit.org/projects/multiverse-inventories/).

Add all the worlds to a "share" with the `last_location` property set. Or add `last_location` to the `use_optionals` list in the config.

benwoo1110:
Didn't read fully, but this looks like a feature of Multiverse-Inventories.

benwoo1110:
outdated PR

PR #1795, opened 2017-02-13
Fixes IllegalArgumentException on startup

BoxCodes:
Error can be found here: https://hastebin.com/osawetafet.css


dumptruckman:
This seems to imply you have a world without a name. Is that correct?

BoxCodes:
Seems like the server owners set the level-name to nothing. Sorry about this pr, thought it was something to do with Multiverse

PR #1808, opened 2017-04-05
Remove relative paths from potentially imported worlds.

RezzedUp:
Before, relative paths that left the server's root directory were accepted by `/mv import`. Now, the world name supplied by the user has all relative-path components trimmed off.


RezzedUp:
There's an exploit going around where users who gain access to `/mv import` (through other means) will import directories outside of the server's root using relative paths and delete them with `/mv delete`. 

This *should* prevent that from happening (leaving the server's root directory, that is).

MrDienns:
+1 for this merge request.

BrettplayMC:
I support this merge (as well)!

PR #1866, opened 2017-09-20
Add ability to define coordinates in commands (set anchor and spawn)

Phoenix616:
This adds the ability to define the world spawn and anchors via the command and with that from the console or command blocks.

The location string uses the same format as the stored locations to not write code duplicate to the one already present in the SimpleLocationManipulation util. The resulting commands look like this:
`/mv anchor {anchorname} world:x,y,z[:pitch[:yaw]]`
`/mv setspawn world:x,y,z[:pitch[:yaw]]`
`/mv modify setspawn world:x,y,z[:pitch[:yaw]]`


dumptruckman:
This ought to use destination parsing instead since there is an "exact" destination with `e:`.

benwoo1110:
Will be doing this in ACF Command rework. Regardless, thank you for the PR, really appreciate it!

PR #1933, opened 2018-09-09
Option to disable the automatic purge of entities

Brokkonaut:
Sometimes you want mobs not to spawn from the minecraft mob spawning system but you do not want custom spawned mobs to be removed - this config option allows that.


FernFerret:
Yo @dumptruckman you around? I peeked at this PR and it looks pretty simple. Any chance you could review it? I don't have a consistent Linux dev environment atm as I keep tearing it down and rebuilding it with Foreman...

@Brokkonaut I personally think this is justified and it gets my vote, we just have to find some time to get it in.

mibby:
@dumptruckman Any chances on this getting merged now that new commits are coming in?

PR #1946, opened 2018-11-20
Update item economy to use material names instead of item IDs for MC 1.13.2

steve4744:
Update for 1.13.2 

- add `api-version: 1.13` to plugin.yml
- change the item economy currency from int to string (i.e. instead of  item ID, now specify the material name as the currency)
- new 1.13 solid blocks, like coral and stripped wood, are now recognised as safe locations for mvtp (instead of being seen as unsafe LEGACY_AIR blocks)
- use XMaterial utility class by Hex_27 for backward compatibility (https://www.spigotmc.org/threads/1-8-to-1-13-itemstack-material-version-support.329630)
- add magma block below player as an unsafe location for mvtp (like fire and lava)
- make carpet a safe block for mvtp (like snow and pressure plates)
- fix the reward message bug saying "charged" instead of "given" when joining a world with negative entry fee 

Tested on 1.9.2, 1.12.2 and 13.2.


dumptruckman:
I appreciate your effort with this, however, I have (finally) implemented a fix using Material rather than String.

PR #2023, opened 2019-06-19
Add new command to tp nearest player from a position like this /mvtp …

elessday:
Add new command to tp nearest player from a position like this /mvtp x y z World


dumptruckman:
This is a neat idea but it would fit a lot better if you designed it as a custom destination.

benwoo1110:
outdated PR

PR #2066, opened 2019-08-07
Fix for worlds not cloning

Destrian:


dumptruckman:
Fixed this in the latest build. Thank you.

Issue #2102, opened 2019-09-23
Anchors are saved locale-sensitively and fail to parse on restart


blubberdiblub:
## Symptom

The anchors got saved as this (note the commas instead of periods):

```yaml
anchors:
  farmdesert: world3:2205,98,70,00,5011,16:256,04:8,80
  farmswamp: world3:-5015,14,67,00,-3982,32:164,06:-0,15
  farmterracotta: world3:4345,02,76,00,7040,28:183,78:13,61
```

Which caused them to fail loading:
```
[00:12:44] [Server thread/WARN]: [Multiverse-Core] The location for anchor 'farmdesert' is INVALID.
[00:12:44] [Server thread/WARN]: [Multiverse-Core] The location for anchor 'farmswamp' is INVALID.
[00:12:44] [Server thread/WARN]: [Multiverse-Core] The location for anchor 'farmterracotta' is INVALID.
```

## Problem description

Part of my locale settings (most notably `LC_CTYPE`, which triggers the problem, but not `LC_NUMERIC`) are on a German locale, i.e. `de_DE.UTF-8`.

This causes java to format numbers with a comma as decimal separator by default when using `%.2f`.

The thing is Multiverse shouldn't be using locale-sensitive formatting when trying to save a machine-readable representation of locations (happens for anchors only, as far as I can tell).

## Relevant code lines

The problematic formatting happens here:
https://github.com/Multiverse/Multiverse-Core/blob/b4b0940918ce49af28bcb11c8b6ce697dc3947ef/src/main/java/com/onarandombox/MultiverseCore/utils/SimpleLocationManipulation.java#L55

And it gets called from this AnchorManager code (which is relevant for this issue):
https://github.com/Multiverse/Multiverse-Core/blob/b4b0940918ce49af28bcb11c8b6ce697dc3947ef/src/main/java/com/onarandombox/MultiverseCore/utils/AnchorManager.java#L115

It also gets called from those 2 locations, but they're not relevant for parsing, as the result is just logged:
https://github.com/Multiverse/Multiverse-Core/blob/b4b0940918ce49af28bcb11c8b6ce697dc3947ef/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java#L442
https://github.com/Multiverse/Multiverse-Core/blob/b4b0940918ce49af28bcb11c8b6ce697dc3947ef/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java#L450

[...]

PR #2103, opened 2019-09-23
Ensure anchors are saved in a machine-readable way

blubberdiblub:
`locationToString()` is primarily used by the AnchorManager to persist
anchors to disk (the other use is for logging).

In a locale that uses periods as decimal separator, this works fine and
the anchors can be loaded when the server restarts.

However, in a locale that doesn't use periods (but commas, for instance)
this produces an `anchors.yml` that cannot be parsed when loaded.

Tying the string formatting in `locationToString()` to an English locale
makes it behave as expected, regardless of the external locale setting.

Fixes: #2102


blubberdiblub:
Looks like your CI setup doesn't like oraclejdk8 anymore. Something like this here: https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766

nicegamer7 (approved):
LGTM!

dumptruckman (approved):
Nice and simple. Sorry for the delay!

PR #2126, opened 2019-11-06
Fix Material error on 1.8

Drc-DEV:


nicegamer7:
We have a separate branch for 1.8 where this is already fixed. Plus, I don't think this is a great solution, given that if the names were to ever change again, it would still build fine without any of us noticing.

Thanks for taking the time to contribute though! If you think what I've said has solved the problem, feel free to close the PR. If not, maybe we can discuss it.

dumptruckman:
Closing this as will not implement.

PR #2168, opened 2020-02-02
Update plugin.yml

Tarluin:
Grouped redundancies as aliases, added permissions for each command and set the default to op for each.

Added multiverse.core.use as permission to access default mv command.

This will prevent players from seeing commands they don't have access to in tab-lists and allow other plugins to get a list of permissions.


benwoo1110:
Thanks for the pr but it has already been done in #2420 and perms are registered with code.

PR #2195, opened 2020-03-21
Update to 1.15.2

jason-s-yu:
Update dependencies to use 1.15.2, including updating the spigot dependency and Vault dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,20 +14,17 @@
     <repositories>
       <!-- Spigot -->
         <repository>
-            <id>spigot-repo</id>
+            <id>spigotmc-repo</id>
             <url>https://hub.spigotmc.org/nexus/content/repositories/snapshots/</url>
         </repository>
         <repository>
             <id>onarandombox</id>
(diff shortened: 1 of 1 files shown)


willies952002:
Simply updating the dependencies to a newer version isn't enough in the case of a 1.15.2 update.

Since `org.bukkit.TravelAgent` was removed recently, `c.o.M.l.MVPlayerListener` and `c.o.M.u.BukkitTravelAgent` fail to compile. These issues would need to be resolved before a release with full 1.15.2 support is possible.

filips123:
> These issues would need to be resolved before a release with full 1.15.2 support is possible.

What does this mean? Does Multiverse already partly works with Minecraft 1.15.2 (maybe with some missing features) or is it completely impossible to use it with 1.15.2?

dumptruckman:
@filips123 It does work with 1.15.2.
@jason-s-yu I'd rather not switch to spigot-api because we do not need any of the additional features it adds.

nicegamer7:
I'm going to close this in favour of #2414.

PR #2208, opened 2020-03-31
Fix Travis Builds for Pull Requests

nicegamer7:
This should fix Travis Builds for Pull Requests. Currently, builds fail since Travis switched to Ubuntu Xenial where Oracle JDK 8 isn't available.

See https://travis-ci.community/t/install-of-oracle-jdk8-is-failing/4365/2.


nicegamer7:
I also removed the sed command as it was unnecessary and causing the build to stall.

PR #2227, opened 2020-04-16
Fix NPE for invalid world argument in gamerule cmd

A248:
Stumbled upon an NPE when using /mv gamerule. I realised it was because I had specified a nonexistent world. This simple PR should fix it.


dumptruckman:
Seems good. Thanks!

PR #2242, opened 2020-04-30
Fixes alias not clearing after cloning the world

wellnesscookie:
This will rather set an alias to an empty string which will automatically be resulted in alias being as same as the name of the cloned world.
Previously, it was not working as intended and kept the alias of the oldWorld.


dumptruckman:
Seems harmless enough...

Issue #2000, opened 2019-05-21
Deprecated EntityCreatePortalEvent


SlimeDog:
Spigot 1.14.2 (Jenkins 2323) git-Spigot-03bd4b0-342194e (MC: 1.14.1) (Implementing API version 1.14.1-R0.1-SNAPSHOT)
WorldGuard 7.0.0-SNAPSHOT;1889-2b854f0
Multiverse-Core 3.0.0-SNAPSHOT-b761

On server start-up:
```
[07:34:59] [Server thread/WARN]: [Multiverse-Core] "Multiverse-Core v3.0.0-SNAPSHOT-b761" has registered a listener for org.bukkit.event.entity.EntityCreatePortalEvent on method "public void com.onarandombox.MultiverseCore.listeners.MVPortalListener.entityPortalCreate(org.bukkit.event.entity.EntityCreatePortalEvent)", but the event is Deprecated. "Server performance will be affected"; please notify the authors [Rigby, fernferret, lithium3141, main--, dumptruckman].
```

MonstahGames:
Yep, having this too.

0xC1A551C:
same

SlimeDog:
It's universal. No reason to keep adding "me too". :)

GCD56:
Tried the latest available build: here is the full text from the Paper Server startup...
(Posted as reference to others looking to post this issue, not to bug the developers)
Multiverse-Core v4.0.0-SNAPSHOT-b764" has registered a listener for org.bukkit.event.entity.EntityCreatePortalEvent on method "public void com.onarandombox.MultiverseCore.listeners.MVPortalListener.entityPortalCreate(org.bukkit.event.entity.EntityCreatePortalEvent)", but the event is Deprecated. "Server performance will be affected"; please notify the authors [Rigby, fernferret, lithium3141, main--, dumptruckman]

Jadiction:
I noticed I forgot to inform anyone viewing this that Multiverse is aware of the deprecated event that you're informed about in console. The "performance impact" isn't noticeable and is fine to ignore.

Andre601:
Since this seems to be a known thing I guess I don't have to mention that I have this too?

SlimeDog:
Nope. You didn't have to mention it. :)

GCD56:
[...]

PR #2261, opened 2020-05-17
Remove Deprecated Portal Event

nicegamer7:
This PR removes the EntityCreatePortalEvent listener, thus closing #2000.

This does not affect the usage of portalForm, and was tested on 1.13.2, 1.14.4, 1.15.2.


nicegamer7:
I'm going to close this since there's already a branch that covers this (it's also probably not merged because of something).

Issue #2278, opened 2020-06-06
The Bitly Link Shortener Doesn't Work


nicegamer7:
This looks like a simple fix. BitlyURLShortener throws an IOException if the resultant link doesn't begin with `http://j.mp/`, this just needs to be changed to `https://j.mp/`.

Issue #2277, opened 2020-06-06
Pasting to Hastebin is broken


nicegamer7:
Pasting to Hastebin currently errors.

```
[21:01:52 WARN]: com.onarandombox.MultiverseCore.utils.webpaste.PasteFailedException: java.io.IOException: Server returned HTTP response code: 403 for URL: https://hastebin.com/documents
[21:01:52 WARN]:        at com.onarandombox.MultiverseCore.utils.webpaste.HastebinPasteService.postData(HastebinPasteService.java:62)
[21:01:52 WARN]:        at com.onarandombox.MultiverseCore.commands.VersionCommand.postToService(VersionCommand.java:213)
[21:01:52 WARN]:        at com.onarandombox.MultiverseCore.commands.VersionCommand.access$000(VersionCommand.java:40)
[21:01:52 WARN]:        at com.onarandombox.MultiverseCore.commands.VersionCommand$1.run(VersionCommand.java:179)
[21:01:52 WARN]:        at org.bukkit.craftbukkit.v1_15_R1.scheduler.CraftTask.run(CraftTask.java:84)
[21:01:52 WARN]:        at org.bukkit.craftbukkit.v1_15_R1.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:54)
[21:01:52 WARN]:        at com.destroystokyo.paper.ServerSchedulerReportingWrapper.run(ServerSchedulerReportingWrapper.java:22)
[21:01:52 WARN]:        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[21:01:52 WARN]:        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[21:01:52 WARN]:        at java.lang.Thread.run(Thread.java:748)
[21:01:52 WARN]: Caused by: java.io.IOException: Server returned HTTP response code: 403 for URL: https://hastebin.com/documents
[21:01:52 WARN]:        at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1900)
[21:01:52 WARN]:        at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1498)
[21:01:52 WARN]:        at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:268)
[...]

Issue #2219, opened 2020-04-09
Multiverse wont let my server start on java 14


lukeeexd:
Hi, 

Using java 14 the server starts up then stops. I found out through binary search that the issue is with multiverse. I deleted multiverse folder and it started fine, however when I imported a world it crashes, then starting up the server again it boots then shuts down. Here is what it put out https://paste.tws.gg/?08eb8469e5fc67de#CcXb99DwqRVbd3z9BUEykiJJSvmcTtESrRaLavVJptsn

benwoo1110:
The error looks like its from EpicWorldGenerator. Another thing to note is that minecraft only officially support java 8. Although java 14 is backwards compatible, it may not be 100%.

lukeeexd:
> The error looks like its from EpicWorldGenerator. Another thing to note is that minecraft only officially support java 8. Although java 14 is backwards compatible, it may not be 100%.

Hi, thanks for the reply. Java 11,13 and 14 all work with minecraft. Infact ZGC is the best with java 14. I use it on another server, I'll make an issue with EWG too.

benwoo1110:
Yes, as i mentioned Java 11,13 and 14 are all backwards compatible with java 8. But minecraft and plugins for it is coded and built based on java 7/8, and may not compatible with newer versions of Java 11,13 and 14, bcu is highly dependent on how they are programmed. In mv's case (and most plugins), it is only tested for java 8 (sometimes 7) support.

nicegamer7:
Are you still getting the error? If so, can you post your log once again?

nicegamer7:
I'm going to close this due to inactivity and because #2279 was merged. Feel free to open a new issue if needed.

Issue #2215, opened 2020-04-06
Sometimes MV Fails to Enable


MidnightSugar:
When I add in plotsquared worlds, I get this error:

```
>.... [15:31:12 ERROR]: Error occurred while enabling Multiverse-Core v4.1.1-SNAPSHOT-b777 (Is it up to date?)
java.lang.NullPointerException: null
at com.onarandombox.buscript.ScriptManager.<init>(ScriptManager.java:31) ~[?:?]
at com.onarandombox.buscript.Buscript.<init>(Buscript.java:77) ~[?:?]
at com.onarandombox.buscript.Buscript.<init>(Buscript.java:65) ~[?:?]
at com.onarandombox.MultiverseCore.MultiverseCore.initializeBuscript(MultiverseCore.java:353) ~[?:?]
at com.onarandombox.MultiverseCore.MultiverseCore.onEnable(MultiverseCore.java:337) ~[?:?]
at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:263) ~[patched_1.15.2.jar:git-Paper-162]
at org.bukkit.plugin.java.JavaPluginLoader.enablePlugin(JavaPluginLoader.java:337) ~[patched_1.15.2.jar:git-Paper-162]
at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:483) ~[patched_1.15.2.jar:git-Paper-162]
at org.bukkit.craftbukkit.v1_15_R1.CraftServer.enablePlugin(CraftServer.java:470) ~[patched_1.15.2.jar:git-Paper-162]
at org.bukkit.craftbukkit.v1_15_R1.CraftServer.enablePlugins(CraftServer.java:384) ~[patched_1.15.2.jar:git-Paper-162]
at net.minecraft.server.v1_15_R1.MinecraftServer.a(MinecraftServer.java:482) ~[patched_1.15.2.jar:git-Paper-162]
at net.minecraft.server.v1_15_R1.DedicatedServer.init(DedicatedServer.java:298) ~[patched_1.15.2.jar:git-Paper-162]
at net.minecraft.server.v1_15_R1.MinecraftServer.run(MinecraftServer.java:885) ~[patched_1.15.2.jar:git-Paper-162]
at java.lang.Thread.run(Thread.java:748) [?:1.8.0_212]
[...]

PR #2279, opened 2020-06-07
Update GameMode command and increase Java support.

nicegamer7:
This PR updates the Gamerule, Gamerules, and Version commands.

The Gamerule command no longer uses deprecated methods.

The Gamerules command had an NPE fixed and no longer uses deprecated methods.

The Version command had the Bitly Link Shortener fixed (#2278), pasting to Hastebin (#2277), and some general cleanup.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
@@ -350,9 +350,14 @@ public class MultiverseCore extends JavaPlugin implements MVPlugin, Core {
      * Initializes the buscript javascript library.
      */
     private void initializeBuscript() {
-        buscript = new Buscript(this);
-        // Add global variable "multiverse" to javascript environment
-        buscript.setScriptVariable("multiverse", this);
+        try {
+            buscript = new Buscript(this);
+            // Add global variable "multiverse" to javascript environment
+            buscript.setScriptVariable("multiverse", this);
+        } catch (NullPointerException e) {
+            buscript = null;
+            Logging.warning("Buscript failed to load! The script command will be disabled!");
+        }
     }
 
     private void initializeDestinationFactory() {
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
@@ -25,7 +25,7 @@ public class GameruleCommand extends MultiverseCommand {
 
     public GameruleCommand(MultiverseCore plugin) {
         super(plugin);
-        this.setName("Set a Minecraft gamerule for a World.");
+        this.setName("Set a Minecraft Game Rule for a World.");
         this.setCommandUsage("/mv gamerule " + ChatColor.GREEN + "{RULE} {VALUE}" + ChatColor.GOLD + " [WORLD]");
         this.setArgRange(2, 3);
         this.addKey("mv gamerule");
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
@@ -10,6 +10,7 @@ package com.onarandombox.MultiverseCore.commands;
 import com.onarandombox.MultiverseCore.MultiverseCore;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
+import org.bukkit.GameRule;
 import org.bukkit.World;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
@@ -24,7 +25,7 @@ public class GameruleCommand extends MultiverseCommand {
 
     public GameruleCommand(MultiverseCore plugin) {
         super(plugin);
-        this.setName("Set a Minecraft Game Rule for a World.");
+        this.setName("Set a Minecraft gamerule for a World.");
         this.setCommandUsage("/mv gamerule " + ChatColor.GREEN + "{RULE} {VALUE}" + ChatColor.GOLD + " [WORLD]");
         this.setArgRange(2, 3);
         this.addKey("mv gamerule");
@@ -55,7 +56,7 @@ public class GameruleCommand extends MultiverseCommand {
             return;
         }
 
-        final String gameRule = args.get(0);
+        final GameRule gameRule = GameRule.getByName(args.get(0));
         final String value = args.get(1);
         final World world;
         if (args.size() == 2) {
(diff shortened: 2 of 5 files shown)


nicegamer7:
Hey @main--, I sent you an email! This is just verification as explained in my email.

nicegamer7:
I just pushed a commit that fixes (or rather, handles) #2215 and possibly #2219.

nicegamer7:
While working on the webpaste package, I realized that all of the PasteServices had the same method used to upload data. So I decided to move that method to HttpAPIClient. One thing led to another, and I decided to make it support the use of auth tokens so that we can support the new Bitly V4 API and pasting to Gist again!

Now, PasteService and URLShortener inherit from HttpAPIClient. I also added to our `pom.xml` so that we don't have to have our Bitly token on GitHub. I tested out all the pasting services (including Gist) and they're all working as expected. The behaviour when there is no GitHub token (which is the default, unless a user compiles MV with one) is to ignore the pasting flag.

I realize that was a rather large commit, I guess I got carried away 😋.

nicegamer7:
I think this PR has gotten rather large... it's now ready for review!

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -350,9 +350,14 @@ public void onEnable() {
...
-        // Add global variable "multiverse" to javascript environment
-        buscript.setScriptVariable("multiverse", this);
+        try {
+            buscript = new Buscript(this);
+            // Add global variable "multiverse" to javascript environment
+            buscript.setScriptVariable("multiverse", this);
+        } catch (NullPointerException e) {
+            buscript = null;
+            Logging.warning("Buscript failed to load! The script command will be disabled!");
+        }
```
dumptruckman, line 360: This could have implications for MVPortals.

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
```diff
@@ -64,17 +65,38 @@ public void runCommand(CommandSender sender, List<String> args) {
...
+                    + " is not a valid gamerule.");
         } else {
-            sender.sendMessage(ChatColor.RED + "Failure!" + ChatColor.WHITE + " Gamerule " + ChatColor.AQUA + gameRule
-                    + ChatColor.WHITE + " cannot be set to " + ChatColor.RED + value);
+            boolean success = false;
+
+            try {
+                if (gameRule.getType() == Boolean.class) {
+                    if (value.equalsIgnoreCase("true")) success = world.setGameRule(gameRule, true);
+                    else if (value.equalsIgnoreCase("false")) success = world.setGameRule(gameRule, false);
```
dumptruckman, line 82: Not a big fan of single line if statements unless it is only if (no else) and is VERY simple. Definitely no assignment. Please use braces here. Also, looks like we should probably consider the `else` case here too.

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
```diff
@@ -64,17 +65,38 @@ public void runCommand(CommandSender sender, List<String> args) {
...
+                } else if (gameRule.getType() == Integer.class) {
+                    success = world.setGameRule(gameRule, Integer.parseInt(value));
+                } else {
+                    sender.sendMessage(ChatColor.RED + "Failure!" + ChatColor.WHITE + " Gamerule " + ChatColor.AQUA + gameRule.getName()
+                            + ChatColor.WHITE + " isn't supported yet, please let us know about it.");
+                    return;
+                }
+            } catch (NumberFormatException ignored) {}
```
dumptruckman, line 90: Can we restrict the scope of this NumberFormatException try/catch to the part that actually needs it? (line 84). Also, we should probably display an error to the user if it happens.

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
```diff
@@ -64,17 +65,38 @@ public void runCommand(CommandSender sender, List<String> args) {
...
-        if (world.setGameRuleValue(gameRule, value)) {
-            sender.sendMessage(ChatColor.GREEN + "Success!" + ChatColor.WHITE + " Gamerule " + ChatColor.AQUA + gameRule
-                    + ChatColor.WHITE + " was set to " + ChatColor.GREEN + value);
+        if (gameRule == null) {
+            sender.sendMessage(ChatColor.RED + "Failure! " + ChatColor.AQUA + args.get(0) + ChatColor.WHITE
+                    + " is not a valid gamerule.");
         } else {
-            sender.sendMessage(ChatColor.RED + "Failure!" + ChatColor.WHITE + " Gamerule " + ChatColor.AQUA + gameRule
-                    + ChatColor.WHITE + " cannot be set to " + ChatColor.RED + value);
+            boolean success = false;
```
dumptruckman, line 77: I'm not keen on this kind of flag setting pattern for something like this. Usually the following can be written in such a way that you don't need to track a flag like this through the code. Rationale: It can be easy to lose track of what happened to this flag by the end of the following code block when the flag is read and acted on. It would be more clear what was happening if the success condition was obvious when it was needed to be known. I can go into more detail about this in discord if this explanation isn't clear enough.

src/main/java/com/onarandombox/MultiverseCore/commands/GamerulesCommand.java
```diff
@@ -60,6 +61,11 @@ public void runCommand(CommandSender sender, List<String> args) {
             world = p.getWorld();
         } else {
             world = Bukkit.getWorld(args.get(0));
+            if (world == null) {
+                sender.sendMessage(ChatColor.RED + "Failure!" + ChatColor.WHITE + " World " + ChatColor.AQUA + args.get(0)
+                        + ChatColor.WHITE + " does not exist.");
+                return;
+            }
```
dumptruckman, line 68: Nice catch here.

src/main/java/com/onarandombox/MultiverseCore/commands/VersionCommand.java
```diff
@@ -35,94 +39,89 @@
...
+        StringBuilder result;
         try {
             FileReader reader = new FileReader(filename);
             BufferedReader bufferedReader = new BufferedReader(reader);
             String line;
-            result = "";
+            result = new StringBuilder();
             while ((line = bufferedReader.readLine()) != null) {
-                result += line + '\n';
+                result.append(line).append(System.lineSeparator());
```
dumptruckman, line 113: If this is the data going to the paste service I would think `\n` would be appropriate, no?

src/main/java/com/onarandombox/MultiverseCore/commands/GamerulesCommand.java
nicegamer7: Thanks 😄

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
nicegamer7: Absolutely, will get braces in there. Also regarding the `else` case, see below on the flag.

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
nicegamer7: Sounds good, will be done. As for why no error is displayed, that's handled right after (see below).

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
nicegamer7: I agree flags are not the greatest way to get things done. However, in an attempt to reduce duplicate code and multiple exit points, I couldn't think of another way.

Regarding both the above reviews, since the `success` flag is set to `false` at the beginning, there is no need to display to check if anything other than `true` or `false` was entered, nor is there a need to display an error when a number is not entered.

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
nicegamer7: The issue that this addresses is rather obscure, I was hoping you could take a look at it.

Anyways, if you'd like to see this in MV Portals, let me know.

src/main/java/com/onarandombox/MultiverseCore/commands/VersionCommand.java
nicegamer7: In this case, yes, I don't see anything wrong with `\n`. I was mainly focused on switching to the system's line ending because the info is also printed to the console/logs, but since this is only used for pasting, it might be better to stick with `\n`.

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
dumptruckman: I just mean that MV Portals probably expects this to not ever be null.

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
dumptruckman: Right, but couldn't the message here indicate that true or false is expected?

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
nicegamer7: I checked, and you're right. Would you like me to open a PR to take care of it?

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
nicegamer7: Done.

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
nicegamer7: Done, and added a more descriptive error message.

src/main/java/com/onarandombox/MultiverseCore/commands/GameruleCommand.java
nicegamer7: I've removed the flag and made the error messages more descriptive.

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
dumptruckman: Sure. Did you do this already?

nicegamer7: Not yet, doing it now.

nicegamer7:
I've rebased this PR and opened another for MV Portals to accommodate the change made to Buscript.

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
dumptruckman: We still need a fix for Portals to deal with this.

src/main/java/com/onarandombox/MultiverseCore/commands/GamerulesCommand.java
```diff
@@ -68,7 +74,7 @@ public void runCommand(CommandSender sender, List<String> args) {
                 gameRules.append(ChatColor.WHITE).append(", ");
             }
             gameRules.append(ChatColor.AQUA).append(gameRule).append(ChatColor.WHITE).append(": ");
-            gameRules.append(ChatColor.GREEN).append(world.getGameRuleValue(gameRule));
+            gameRules.append(ChatColor.GREEN).append(world.getGameRuleValue(GameRule.getByName(gameRule)));
```
dumptruckman, line 77: This could potentially cause issues if the enum values ever change as we'll have persisted them differently. It's an unlikely thing but something to consider.

dumptruckman (approved):

src/main/java/com/onarandombox/MultiverseCore/commands/GamerulesCommand.java
nicegamer7: I don't understand the concern here, as `gameRule` comes from the Bukkit World's list of gamerules. Could you clarify?

dumptruckman: Ah, nevermind then :)

Issue #1918, opened 2018-06-30
Switch from mcstats to bstats


garypwn:
Is far as I can tell, mcstats has not been supported since 2015. The site is still up, but there have been significant patches of downtime, the creator does not respond to issues on his github, and most plugin authors have bailed and switched to bStats. Multiverse has mcstats as a compile dependency, and it makes building any project that depends on Multiverse a bad time.

FernFerret:
Ah interesting, sorry, this one slipped through my radar, I'm going through trying to do a bit of spring (winter?) cleaning. This one is totally on my list now.

Issue #2163, opened 2020-01-27
Missing dependency declaration


SlimeDog:
Spigot 1.15.2 (Jenkins 2596)
Multiverse-Core 4.1.1-b777

**Description**
Beginning with Spigot 1.15.2 (Jenkins 2596), missing dependency declarations are reported in the log on server start-up as WARN messages.

Reference: [https://hub.spigotmc.org/jira/browse/SPIGOT-5546](https://hub.spigotmc.org/jira/browse/SPIGOT-5546)

The relevant log entry is:
```
[22:03:37] [Server thread/WARN]: [Multiverse-Core] Loaded class net.milkbowl.vault.economy.Economy from Vault v1.7.3-b${env.TRAVIS_BUILD_NUMBER} which is not a depend, softdepend or loadbefore of this plugin.
```

**Steps to reproduce the behavior**
- Update Spigot to 1.15.2 latest
- Start the server
- Search the log for 'Loaded class'

**Expected behavior**
Dependencies should be declared in plugin.yml

andrewkm:
Same here.

`[Multiverse-Core] Loaded class net.milkbowl.vault.economy.Economy from Vault v1.7.3-b${env.TRAVIS_BUILD_NUMBER} which is not a depend, softdepend or loadbefore of this plugin.`

bonjourkemal:
I'm following.

nicegamer7:
It seems we forgot to close this. This issue is fixed in our latest CI build.

SlimeDog:
Please check the CI server (http://ci.onarandombox.com/view/Multiverse/job/Multiverse-Core/). It's been inaccessible (504 gateway error) for days.

nicegamer7:
Thanks for letting us know, we are aware of the issue.

PR #2280, opened 2020-06-07
Use bStats

### LINKED ISSUE #2252: Async tasks error witch PaperMC #280
Hello,

I was on PaperMC 273 1.15.2 without any issue, 7 worlds with Multiverse-core. Today, i upgraded to PaperMC 280 and when my server close, i have this error every time. I reported it to PaperMC, they told me to see with Multiverse.

[10:40:53] [Server thread/ERROR]: Nag author: 'Rigby' of 'Multiverse-Core' about the following: This plugin is not properly shutting down its async tasks when it is being shut down. This task may throw errors during the final shutdown logs and might not complete before process dies.

Thank you.

nicegamer7:
This PR switches our metrics system from MCStats to bStats. In the process we lost one metric (Worlds vs Loaded Worlds) because bStats currently does not support Multi-Line Charts.

Fixes #2252, implements #1918.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,10 +33,9 @@
             <id>minebench-repo</id>
             <url>https://repo.minebench.de/</url>
         </repository>
-        <!-- Has a copy of metrics R8-SNAPSHOT !-->
         <repository>
-            <id>elmakers-repo</id>
-            <url>http://maven.elmakers.com/repository/</url>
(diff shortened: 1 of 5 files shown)


nicegamer7:
I just added Vault as a softdepend, fixing #2163. This is now ready for review!

PR #2353, opened 2020-07-18
Fix creation and import Exploit

xSavior-of-God:
Through this fix you can avoid deleting of important folders.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
@@ -46,10 +46,15 @@ public class CreateCommand extends MultiverseCommand {
         this.addCommandExample("/mv create " + ChatColor.GOLD + "moonworld" + ChatColor.GREEN + " normal" + ChatColor.DARK_AQUA + " -g BukkitFullOfMoon");
         this.worldManager = this.plugin.getMVWorldManager();
     }
-
+	
+	private String trimWorldName(String userInput) {
+        // Removes relative paths.
+        return userInput.replaceAll("^[./\\\\]+", "");
+    }
+	
     @Override
     public void runCommand(CommandSender sender, List<String> args) {
-        String worldName = args.get(0);
+        String worldName = trimWorldName(args.get(0));
         File worldFile = new File(this.plugin.getServer().getWorldContainer(), worldName);
         String env = args.get(1);
         String seed = CommandHandler.getFlag("-s", args);
@@ -66,7 +71,13 @@ public class CreateCommand extends MultiverseCommand {
                 useSpawnAdjust = false;
             }
         }
-
+		
+		// Make sure the world name doesn't contain the words 'plugins' and '.dat'
+		if(worldName.contains("plugins")||worldName.contains(".dat")){
+			sender.sendMessage(ChatColor.RED + "Multiverse cannot create a world that contains 'plugins' or '.dat'");
+            return;
+		}
+		
         if (this.worldManager.isMVWorld(worldName)) {
             sender.sendMessage(ChatColor.RED + "Multiverse cannot create " + ChatColor.GOLD + ChatColor.UNDERLINE
                     + "another" + ChatColor.RESET + ChatColor.RED + " world named " + worldName);
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
@@ -121,6 +121,12 @@ public class ImportCommand extends MultiverseCommand {
             this.showHelp(sender);
             return;
         }
+		
+		// Make sure the world name doesn't contain the words 'plugins' and '.dat'
+		if(worldName.contains("plugins")||worldName.contains(".dat")){
+			sender.sendMessage(ChatColor.RED + "Multiverse cannot create a world that contains 'plugins' or '.dat'");
+            return;
+		}
 
         // Make sure we don't already know about this world.
         if (this.worldManager.isMVWorld(worldName)) {


src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -46,10 +46,15 @@ public CreateCommand(MultiverseCore plugin) {
         this.addCommandExample("/mv create " + ChatColor.GOLD + "moonworld" + ChatColor.GREEN + " normal" + ChatColor.DARK_AQUA + " -g BukkitFullOfMoon");
         this.worldManager = this.plugin.getMVWorldManager();
     }
-
+	
+	private String trimWorldName(String userInput) {
+        // Removes relative paths.
+        return userInput.replaceAll("^[./\\\\]+", "");
+    }
```
nicegamer7, line 53: Why wouldn't `\\` suffice (in place of `\\\\`)?

Also, please use spaces for indenting.

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -66,7 +71,13 @@ public void runCommand(CommandSender sender, List<String> args) {
                 useSpawnAdjust = false;
             }
         }
-
+		
+		// Make sure the world name doesn't contain the words 'plugins' and '.dat'
+		if(worldName.contains("plugins")||worldName.contains(".dat")){
+			sender.sendMessage(ChatColor.RED + "Multiverse cannot create a world that contains 'plugins' or '.dat'");
+            return;
+		}
```
nicegamer7, line 79: There is no need for this check, since MV will not create a world if the folder already exists.

In this case, if MV is installed, the plugins folder already exists. Also, I don't understand why we would need to check for `.dat`.

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
```diff
@@ -121,6 +121,12 @@ public void runCommand(CommandSender sender, List<String> args) {
             this.showHelp(sender);
             return;
         }
+		
+		// Make sure the world name doesn't contain the words 'plugins' and '.dat'
+		if(worldName.contains("plugins")||worldName.contains(".dat")){
+			sender.sendMessage(ChatColor.RED + "Multiverse cannot create a world that contains 'plugins' or '.dat'");
+            return;
+		}
```
nicegamer7, line 129: Same comment as in the Create Command.

Also, I think the import command could benefit from the trim method you added to the Create Command.

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
xSavior-of-God: I copied the function present in 
https://github.com/Multiverse/Multiverse-Core/blob/master/src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java#L99

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
xSavior-of-God: if you create a world inside the plugins folder with name test.dat, and then try to import that world by doing '/mv import ./plugins/' MV will import it because there is a file/folder with extension ".dat", so to avoid this problem would be enough to forbid to insert of the word ".dat" in the name of the world.

Example of this exploit: "/mv create /plugins/test.dat/world/"

Thinking about it, in this case the ".dat" check would be enough.

```
if( worldName.contains(".dat") ) {
         sender.sendMessage( ChatColor.RED + "Multiverse cannot create a world that contains '.dat'" );
         return;
}
```

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
xSavior-of-God: Here, on the other hand, I would like to prevent in general from importing worlds with the word "plugins" in the middle, so as to avoid other unwanted exploits directed to this folder.

```
// Make sure the world name doesn't contain the words 'plugins'
if( worldName.contains("plugins") ){
         sender.sendMessage( ChatColor.RED + "Multiverse cannot create a world that contains 'plugins'" );
         return;
}
```

xSavior-of-God:
If there are no other corrections, I will modify the code and repeat the pull request if necessary.

nicegamer7:
I'm not actually authorized to accept PRs, so I'd wait until the main dev gives his thoughts until you redo anything.

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
dumptruckman: Huh, this is a good catch. And yeah, I think just the .dat check would be fine.

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
```diff
@@ -168,4 +174,4 @@ public void runCommand(CommandSender sender, List<String> args) {
                 Command.broadcastCommandMessage(sender, ChatColor.RED + "Failed!");
         }
     }
-}
+}
```
dumptruckman, line 177: We prefer a blank line at the end.

dumptruckman (changes requested):
> If there are no other corrections, I will modify the code and repeat the pull request if necessary.

Sorry for the delay in checking this out. I made some comments so yeah, need a couple changes but otherwise this is great.

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
nicegamer7: Wouldn't it be better to just deny creating worlds inside of other folders?

benwoo1110:
Will be integrating this in ACF Command rework. Not sure if we want to merge it before that, like v4.3?

benwoo1110:
@xSavior-of-God Are you still interested in proceeding on this PR?

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
nicegamer7: I figured out why `\\\\` is needed.

The first `\\` is to escape the regex escape char (`\`). The second `\\` is to escape the backslash.

xSavior-of-God:
Yes, i need this fix!

benwoo1110:
I think I recently added something similar, see https://github.com/Multiverse/Multiverse-Core/blob/main/src/main/java/com/onarandombox/MultiverseCore/utils/WorldNameChecker.java

I suggest you take a look at it and see if it has fixed the exploits you are concern with. I haven't use WorldNameChecker in the create command class, but it's probably easy to implement.

Another thing is I heard of people having sub-folders dir for worlds e.g. `hub/worldname`, since / is a valid char world worldname. Seems like this pr will prevent this behaviour?

xSavior-of-God:
**1 year ago I said how to fix/remedy the problem... and to date it has not been fixed...**

**I also mentioned how to replicate it, which to date may have caused problems for some servers owners... with the hope that it will be fixed as soon as possible, but this was not...**

**So now I tell you how things are...**
---

Did you know that if you want you can create a world inside folders like `/sys/`, `/var/`, `/usr/`, `/root/`, `/home/`, etc...?
_I don't know how safe it is to use Multiverse on a productive server..._ ok it's not recommended to start your server with root user, it's recommended to limit the permissions to the user who start the server. 
Everyone obviously manages as they see fit, but a small and simple control or just a limitation via config (such as: "_Hey by activating this you can risk hack attacks, such as deletion of important folders..._") it could already be convenient, like I just enable it when I need, and I disable it when I don't need it anymore. (also via API if you use mv)

In any case, for example, it would be enough to know the path of the server folder to **completely destroy it** (or just use "`./`" that refer to the root of your server) by creating a world inside it and then deleting it... honestly, I told you how to do it and how to fix it, I think, it should be good a simple check of the name that removes the characters like '`/`', '`\`' and '`.`'...

---
## However the problem still persists!

### How to replicate delete plugins folder?
**Simple just type this 2 commands:**

```
/mv create ./plugins/test.dat/dio/ normal
/mv import ./plugins/ normal
```

**NB**: Do not use contains in reverse!
https://github.com/Multiverse/Multiverse-Core/blob/28cee291c66a9f3691d5e42fc9935a7228337778/src/main/java/com/onarandombox/MultiverseCore/utils/WorldNameChecker.java#L39-L54

do it like this

```JAVA
for(String check : BLACKLIST_NAMES) 
  if (worldName.contains(check))
    return NameStatus.BLACKLISTED;
```

---

I hope you have understood the seriousness of the problem and that I have not offended you (in case sorry), but above all I hope that the problem will be solved as soon as possible!

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
benwoo1110: yea, but can potential break all servers that has worlds in sub directory. bcu you could do hub/worldname, and I think that is valid.

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
benwoo1110: Should we restrict so tightly tho. I get that world shouldn’t be in plugin directory, but doesn’t means a worldname can’t contain plugin, e.g. plugin_world can be valid.

benwoo1110:
ideally, we should check based on directory instead of just removing all /, possibly:
1. target world folder is within the world container path, not /home /root as you said.
2. and also not sub directory path of certain blacklisted folder, e.g. plugins/ logs/

so that we control, but not straight up deny worlds in sub directory, to prevent breaking changes while fixing this exploit.

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
xSavior-of-God: You could just add a (configurable) check to disable dots and slashes... I think that's enough

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
xSavior-of-God: As I said earlier...
_You could just add a (configurable) check to disable dots and slashes... I think that's enough_

benwoo1110 (approved):
I will merge it as it is, and look into configurable options, as well as reducing false positives as I mentioned in the near future.

xSavior-of-God:
Why do you have to get too complicated? 
You can simply make an option in the config that makes it possible to use the '`/`', '`\`' and the '`.`', doing this avoids the biggest problem, then check if it is an important folder like: `logs`, `plugins`, `libs`, etc... (also this last if configurable would be the top)

benwoo1110:
it's not... too complicated, I'm just providing ideas into how we can improve the world name checking system. Yes - things at mv are slow at times, but we do our best with the limited free time we have.

PR #2359, opened 2020-07-21
Use the World Attribute from the Location Class

nicegamer7:
While trying to figure out the problem behind #2332 (it ended up being a Spigot bug), I realized that MV's SpawnLocation Class uses it's own World attribute and ignores the one present in the Location Class.

I couldn't understand why this was done, especially since MVWorld.getSpawnLocation() would return with a location with a null World. This PR removes SpawnLocation's World attribute in favour of World's.


nicegamer7:
I'm marking this as a draft until further notice... maybe indefinitely.

benwoo1110:
Is this.. still needed?

nicegamer7:
It was never needed, it's just something I realized that confused me. The reason I set this PR as a draft is because dump and I didn't know why this was done in the first place, so we weren't sure if this change would break anything. I think this change *might* reduce memory usage by a tiny bit.

benwoo1110:
I believe weak reference is used bcu world properties are kept in memory even if the world is unloaded.

I just updated to `toString` method so debug output doesnt keep showing null world https://github.com/Multiverse/Multiverse-Core/commit/3fc3663d5d2839f2b48df478176244e204165556

PR #2379, opened 2020-08-13
load cloned world with doLoad() instead of addWorld() (fixes #2378)

### LINKED ISSUE #2378: Cloning world initially applies default properties to cloned world
### Steps to reproduce
- Create a world named "original"
- Disallow monsters with `mvm set monsters false original`
- Clone the world with `mv clone original clone`
- Teleport to the clone with `mvtp clone`
- Spawn a zombie with a spawn egg

### Expected behaviour
The zombie does not appear in the world "clone", as the world "original" had monsters set to false.

### Actual behaviour
The zombie does appear in the world "clone".

### Further notes
- Having a look into `worlds.yml` directly after cloning verifies that the default properties are set
- `mvinfo clone` prints the correct properties, but they are actually not applied
- Restarting the server fixes the problem

### Cause
In `WorldManager.java` around line 200 the cloned world is added with `addWorld()`. This creates a set of default properties. After the `addWorld()` call, the original properties are copied to the properties map of the cloned world, but the cloned world does not apply them.

### Fix
I created a patch fixing this problem. Expect the pull request in a couple of minutes.

rti:
This pull request fixes #2378 

We actually can skip all the `WorldCreator` setup done by `addWorld()` as the cloned world is ready and just needs to be loaded. Just using `doLoad()` enables us to cleanly apply the copied properties, which is the intention of this patch.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
@@ -193,16 +193,31 @@ public class WorldManager implements MVWorldManager {
 
         if (newWorldFile.exists()) {
             Logging.fine("Succeeded at copying files");
-            if (this.addWorld(newName, environment, seedString, worldType, generateStructures, generator, useSpawnAdjust)) {
-                // getMVWorld() doesn't actually return an MVWorld
-                Logging.fine("Succeeded at importing world");
-                MVWorld newWorld = (MVWorld) this.getMVWorld(newName);
-                newWorld.copyValues(this.worldsFromTheConfig.get(oldName));
-                // don't keep the alias the same -- that would be useless
-                newWorld.setAlias("");
-                return true;
+
+            // initialize new properties with old ones
+            WorldProperties newProps = new WorldProperties();
+            newProps.copyValues(this.worldsFromTheConfig.get(oldName));
+            // don't keep the alias the same -- that would be useless
+            newProps.setAlias("");
+            // store the new properties in worlds config map
+            this.worldsFromTheConfig.put(newName, newProps);
+
+            // save the worlds config to disk (worlds.yml)
+            if (!saveWorldsConfig()) {
+                this.plugin.log(Level.SEVERE, "Failed to save worlds.yml");
+                return false;
             }
+
+            // actually load the world
+            if (doLoad(newName)) {
+               this.plugin.log(Level.FINE, "Succeeded at loading cloned world '" + newName + "'");
+               return true;
+            }
+
+            this.plugin.log(Level.SEVERE, "Failed to load the cloned world '" + newName + "'");
+            return false;
         }
+
         Logging.warning("Failed to copy files for world '%s', see the log info", newName);
         return false;
     }


src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -193,16 +193,31 @@ public boolean cloneWorld(String oldName, String newName) {
...
+            // initialize new properties with old ones
+            WorldProperties newProps = new WorldProperties();
+            newProps.copyValues(worldsFromTheConfig.get(oldName));
+            // don't keep the alias the same -- that would be useless
+            newProps.setAlias("");
+            // store the new properties in worlds config map
+            worldsFromTheConfig.put(newName, newProps);
+
+            // save the worlds config to disk (worlds.yml)
+            if(!saveWorldsConfig()) {
```
nicegamer7, line 206: Please add a space after `if`.

src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -193,16 +193,31 @@ public boolean cloneWorld(String oldName, String newName) {
...
+                this.plugin.log(Level.SEVERE, "Failed to save worlds.yml");
+                return false;
+            }
+
+            // actually load the world
+            if(doLoad(newName)) {
```
nicegamer7, line 212: Please add a space after `if`.

src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -193,16 +193,31 @@ public boolean cloneWorld(String oldName, String newName) {
...
+               Logging.fine("Succeeded at loading cloned world '" + newName + "'");
```
nicegamer7, line 213: Just for consistency with the the modern MV codebase, this should be changed to from
```java
Logging.fine(...
```
to
```java
this.plugin.log(Level.FINE, ...
```

nicegamer7 (changes requested):
Thanks for the PR! Overall seems good. I just requested some style changes, nothing big.

src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -193,16 +193,31 @@ public boolean cloneWorld(String oldName, String newName) {
...
-                newWorld.setAlias("");
-                return true;
+
+            // initialize new properties with old ones
+            WorldProperties newProps = new WorldProperties();
+            newProps.copyValues(worldsFromTheConfig.get(oldName));
+            // don't keep the alias the same -- that would be useless
+            newProps.setAlias("");
+            // store the new properties in worlds config map
+            worldsFromTheConfig.put(newName, newProps);
```
nicegamer7, line 203: This is also a really small nitpick, but when referring to `worldsFromTheConfig`, please use `this.worldsFromTheConfig` as it makes it easier to see that it's an attribute of the class.

rti:
Thanks a lot for the review. I addressed your comments.

rti:
Not sure whether force-pushing was a good idea here. Github seems to be a bit confused now. Sorry. Hope that works for you.

src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -193,16 +193,31 @@ public boolean cloneWorld(String oldName, String newName) {
...
-                Logging.fine("Succeeded at importing world");
-                MVWorld newWorld = (MVWorld) this.getMVWorld(newName);
-                newWorld.copyValues(this.worldsFromTheConfig.get(oldName));
-                // don't keep the alias the same -- that would be useless
-                newWorld.setAlias("");
-                return true;
+
+            // initialize new properties with old ones
+            WorldProperties newProps = new WorldProperties();
+            newProps.copyValues(worldsFromTheConfig.get(oldName));
```
nicegamer7, line 199: You missed a `this.` over here.

nicegamer7 (changes requested):

nicegamer7:
Force pushing is fine.

src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
rti: Oops. Thanks for pointing that out.

benwoo1110:
Also just a note, usually it's good to create a new branch when making a PR instead of using master.

rti:
You are right. Would have been way cleaner. Though, I created my fork only to create this PR.

rti:
Hm, the travis build stopped while downloading something. Is there a way to re-trigger the build?

benwoo1110 (approved):

nicegamer7 (approved):

nicegamer7:
I figured out how to trigger a restart 👍.

dumptruckman:
Awesome. Thank you for this!

PR #2393, opened 2020-08-27
Disallow world names with spaces

### LINKED ISSUE #2201: Unable to import worlds with spaces
I haven't confirmed this myself.

If I can reproduce this issue, I'll see if I can fix it and open a pull request.

benwoo1110:
This should fix #2201 by giving a more informative message that world name should not have spaces.

A possible improvement is to show the message in-game, but it will also mean I need to add this checking in all the run command methods where world name is concerned.

currently it shows:
image

I also notice when the command hits an error, it just says "Failed" in-game, perhaps its better to change it to "Failed! Check console for errors!". Probably PR for another day haha.


src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -403,6 +418,22 @@ public boolean loadWorld(String name) {
...
 
+    /**
+     * Check if the world name is allowed
+     *
+     * @param name   Name of the world
+     * @return True if the world world name is valid based on regex
+     */
+    private boolean isValidWorldName(String name) {
+        String worldNamePattern = "[a-zA-Z0-9/._-]+";
+        if (!name.matches(worldNamePattern)) {
```
dumptruckman, line 429: I don't see where you were using the imported `Pattern` class but I do tend to prefer to use `Pattern.compile` and store that in a static var. I assume it's faster but I'm not sure I actually know that. :)

dumptruckman (approved):

PR #2395, opened 2020-08-28
Improve chat with RGB support

benwoo1110:
Add support for colour codes directly on `alias`, thus `color` and `style` world properties will not be needed. (But I think we can still keep it unless I can create a conversion or some sort)

Here are some images of how it looks:
cool colour
image


benwoo1110:
I think the regex needs improvement and also should I parse the chat colors for the `prefixchatformat: '[%world%]%chat%'` so that things like the `[` can have colour.

benwoo1110:
Superseded by #2397

PR #2396, opened 2020-08-28
Added ability to add all overworld, nether and end at once

### LINKED ISSUE #2389: Feature Request: optional Import command switch to automatically extract Nether & End dimensions from a Vanilla world save
Problem: I successfully imported a vanilla world save but this only creates an overworld dimension. I want the corresponding Nether and End dimensions.

I tried to manually create folders for these dimensions and move the DIM-1 and DIM1 data to them, but attempting to import them fails because there is no .dat file for them.  I'm not sure how to proceed, do I copy the .dat files from the overworld?

A switch to import Nether and End dimensions from Vanilla world saves would be appreciated.

benwoo1110:
This PR will import the overworld, nether and end altogether. Fixes #2389. 

For now, I added a new command `/mv vanillaimport <worldname>` but maybe it's better to integrate it to the current import command?


nicegamer7:
I was thinking about this a couple days ago and this is the usage I thought would be nice:

```/mvi --nether whatever` or `/mvi -n whatever```

and there would be similar flags for the end, which could be combined to import both. Let me know what you think

benwoo1110:
My thinking is that if someone wants to import a vanilla world they probably want the end and then nether, so I don't think there is a need to split up to different import commands. My new thinking is something like `mv import -v <vanilla_world_name>`.

if they don't want the separate end and nether, they can just `mv import <vanilla_world_name> normal`.

nicegamer7:
I guess that's good too, just FYI, there's a way to check for flags using CommandHandler, but I don't remember how right now. Check the Version command to see how it's done.

benwoo1110:
I will keep the previous implementation and add the new one I said, and then see what dump thinks about it.

As for the flag, from what I can tell its something like `CommandHandler.getFlag("-g", args)`

benwoo1110:
Actually thinking of it again, something like `mv import <vanilla_world_name> --vanilla` with an alias of `mv import <vanilla_world_name> -v` may be more logical so that it aligns with normal import command where the world name comes first.

benwoo1110:
So here is the summary of the changes:

**commands:**
* Added `all` for environment so it imports/create worldname, worldname_nether and worldname_the_end together. Example command: `/mv create WORLDNAME all`
* Added vanilla import, which splits a vanilla world to its respective overworld, nether and end. Example command: `/mv import WORLDNAME vanilla`

**code**:
* Added `splitVanillaWorld()`, `isValidWorld()` and `getPotentialWorlds()` to WorldManager class
* Refractored quite a lot for ImportCommand class
* Changes to CreateCommand to add create `all` feature

nicegamer7:
On Discord yesterday, I was talking with someone who wanted to be able to import or create a "Buffet" world. I looked into it briefly and found it could be done with the Bukkit API.

So do you think it would be a good idea to add it to this PR, or would it be better to separate it? cc: @dumptruckman, @benwoo1110

benwoo1110:
@nicegamer7 I think that is a good feature to add if possible, though it's probably better for another PR, as it will most likely need more time to look into the implementation of it.

benwoo1110:
may split this up to multiple smaller pr

benwoo1110:
Implemented better with #2457

PR #2411, opened 2020-09-19
Webpaste Rework Follow-up

nicegamer7:
This PR is just a follow-up to #2288. It makes a few changes to the code so that it's more in line with MV's coding style.

There's also two miscellaneous changes:
- A new constructor was added to HttpAPIClient, making the code a bit cleaner.
- An IllegalStateException was changed to an IllegalArgumentException because it's more correct.


dumptruckman (approved):
Looks good.

PR #2415, opened 2020-09-26
Remove TravelAgent

nicegamer7:
This PR removes any usage of TravelAgent from MV Core while keeping backwards compatibility.


nicegamer7:
Superseded by #2454.

PR #2420, opened 2020-10-04
Updated plugin.yml commands and some missing alias

benwoo1110:
Not sure why the plugins.yml commands looks so messed up, so here is just a quick fix for it.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/VersionCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/VersionCommand.java
@@ -44,9 +44,10 @@ public class VersionCommand extends MultiverseCommand {
     public VersionCommand(MultiverseCore plugin) {
         super(plugin);
         this.setName("Multiverse Version");
-        this.setCommandUsage("/mv version " + ChatColor.GOLD + "-[bhp] [--include-plugin-list]");
+        this.setCommandUsage("/mv version " + ChatColor.GOLD + "[-b|-h|-p] [--include-plugin-list]");
         this.setArgRange(0, 2);
         this.addKey("mv version");
+        this.addKey("mvver");
         this.addKey("mvv");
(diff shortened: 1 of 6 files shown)


src/main/java/com/onarandombox/MultiverseCore/commands/VersionCommand.java
```diff
@@ -44,9 +44,10 @@
     public VersionCommand(MultiverseCore plugin) {
         super(plugin);
         this.setName("Multiverse Version");
-        this.setCommandUsage("/mv version " + ChatColor.GOLD + "-[bhp] [--include-plugin-list]");
+        this.setCommandUsage("/mv version " + ChatColor.GOLD + "[-b|-h|-p] [--include-plugin-list]");
```
dumptruckman, line 47: This is the standard we use elsewhere?

nicegamer7: I think so.

https://github.com/Multiverse/Multiverse-Core/blob/master/src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java#L34

https://github.com/Multiverse/Multiverse-Core/blob/master/src/main/java/com/onarandombox/MultiverseCore/commands/DebugCommand.java#L26

https://github.com/Multiverse/Multiverse-Core/blob/master/src/main/java/com/onarandombox/MultiverseCore/commands/ModifyCommand.java#L30

benwoo1110: Most other command had the `-` inside the `[]` (so like `[-s]`, tho this is the only instance there is multiple of them. I thought this looks clearer, bcu `-[bph]` could be misunderstood as doing `-bhp` and not `-b` or `-h` or `-p

dumptruckman (approved):

PR #2427, opened 2020-10-06
Adds MVWorldManager#isValidWorld method

benwoo1110:
This is a purpose build method use to check if world folder exist and is valid with level.dat


benwoo1110:

PR #2429, opened 2020-10-06
More noticable warning for deprecated event

benwoo1110:
Add green color to the warning. I hope this will reduce the number of ppl asking about it lol


dumptruckman (approved):
LOVE IT

PR #2438, opened 2020-10-07
Fix config reset

benwoo1110:
Loads the config on plugin disable so config doesnt reset to old values


nicegamer7:
Rather than loading then saving the config files, I think the same thing could be achieved by simply not saving the config files when disabling the plugin.

However, this change might cause issues if something is not yet saved before disabling the plugin. For example, imagine you created a world and for some reason it did not save to the config file. Upon disabling the plugin, the config will not change in either case. So I think we need to make some changes to the SerializableConfig so that any time a change is made it is saved immediately.

benwoo1110:
Closing this bcu we probably need to come up with a better solution

PR #2445, opened 2020-10-09
Fix issue where special chars cause PatternSyntaxException

### LINKED ISSUE #2433: Regex error when running unexpected mv command
### Details
I was **able** to reproduce my issue on a freshly setup and up-to-date server with the latest version of Multiverse plugins with no other plugins and with no kinds of other server or client mods.

**Description**  
https://pastebin.com/raw/hFCMPrB7

**Steps to reproduce**  
run an invalid command like `mv hel[p`

benwoo1110:
Escape regex special chars for help command, fixes #2433

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/PaginatedCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/PaginatedCommand.java
@@ -56,7 +56,7 @@ public abstract class PaginatedCommand<T> extends Command {
      * @return String with regex characters escaped
      */
     protected String cleanFilter(String filter) {
-        return REGEX_SPECIAL_CHARS.matcher(filter.toLowerCase()).replaceAll("\\\\$0");
+        return REGEX_SPECIAL_CHARS.matcher(filter).replaceAll("\\\\$0");
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nicegamer7:
Does this break filtering?

benwoo1110:
From my testing no, anyways it's going to be completely changed in ACF as well.

benwoo1110:
@nicegamer7 can we merge this or smt? (want to delete my old mv forks 😅)

src/main/java/com/onarandombox/MultiverseCore/commands/PaginatedCommand.java
```diff
@@ -40,12 +42,23 @@ protected void setItemsPerPage(int items) {
...
     protected abstract List<T> getFilteredItems(List<T> availableItems, String filter);
 
+    /**
+     * Escape regex special characters from filter
+     *
+     * @param filter The filter-{@link String}.
+     * @return String with regex characters escaped
+     */
+    protected String cleanFilter(String filter) {
+        return REGEX_SPECIAL_CHARS.matcher(filter.toLowerCase()).replaceAll("\\\\$0");
```
nicegamer7, line 59: Any reason you went to lower case here?

benwoo1110: brain goes brrrr

nicegamer7 (approved):
I was originally hesitant to merge this because I wasn't sure if the filter was meant to allow users to search using regex, but I came to the realization that I don't think anyone is searching for a command using regex lol.

benwoo1110:
nah, it was just sole to fix a regex parsing issue

PR #2446, opened 2020-10-09
List available properties for modify commands

benwoo1110:
Add command flag: `/mvm <set|add|remove|clear> -list`, fixes #1399

PR #2447, opened 2020-10-09
Refractor targetWorld for multiverse commands

benwoo1110:
While doing PR #2446 I realised there was quite a lot of copied code for the modify add, clear, remove and set commands, so just refractor it and generalise it to ModifyCommand#getTargetWorld method.

EDIT: I changed it be in MultiverseCommand class so it can be used in other commands such as `/mv gamerule` and `/mv gamerules`


benwoo1110:

PR #2397, opened 2020-08-31
1.16 update

benwoo1110:
### Things changed:
* Bumped Bukkit API used to 1.16.2
* Added `VersionUtils` to allow backwards compatibility when doing version-specific events (e.g. when doing things like checking for respawn anchor)
* Added `FormatUtils` to parse colours (and possibility other formatting stuff in the further)
* Added support for respawn anchors, fix #2329
* Added support for RGB parse for world alias and chat format. implements #2414 
* `MVWorldByAlias` without alias colour codes (So people don't have to type in colour codes for things like `/mvtp` commands)

### Things to note:
* Now uses spigot API as there is no Bukkit API for 1.16
* 1.13.2+ support is maintained
* 1.16.2 does not have `getPortalTravelAgent()` anymore, so code that used that is removed. Though this should not affect anything as code with TravelAgent methods does seem to be in use anymore.
* Respawn anchor charge does not decrease when using with MV-INV bed_spawn sharing. This is due to MV-INV setting the respawn point as bed spawn and the current Bukkit API does not separate bed from anchor.
* Any queries, you can come to our MV discord at https://discord.gg/NZtfKky

### Test build:
* ~[Multiverse-Core-4.1.1-SNAPSHOT.zip](https://github.com/Multiverse/Multiverse-Core/files/5150002/Multiverse-Core-4.1.1-SNAPSHOT.zip)~

PR #2454, opened 2020-10-12
1.15 Update

nicegamer7:
This PR updates the Bukkit dependency to 1.15.2. This is to be followed up by #2397. The reason for splitting it up is just to prevent the 1.16 PR from becoming too large.


nicegamer7:
We'll revisit this after 5.0.0's ACF PRs.

PR #2457, opened 2020-10-14
Adds ability to split and import vanilla world dims.

### LINKED ISSUE #2389: Feature Request: optional Import command switch to automatically extract Nether & End dimensions from a Vanilla world save

benwoo1110:
Fixes #2389.


benwoo1110:
Will re-look at 5.0.0

PR #2487, opened 2020-11-22
Fix teleporting to the calling player's bed

akrantz01:
When using the bed destination, the player's name would always be set causing the `b:playerbed` to not work. This adds an extra check to see if the player's name is `playerbed`. I also added the ability to use `b:my-bed` since it is an invalid player name and will never collide with an actual player' username (assuming Mojang/Microsoft doesn't change the username requirements).

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/destination/BedDestination.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/destination/BedDestination.java
@@ -23,8 +23,8 @@ import org.bukkit.util.Vector;
  * A bed-{@link MVDestination}.
  */
 public class BedDestination implements MVDestination {
-    public static final String CALLER_BED_NAME = "my-bed";
-    public static final String CALLER_BED_STRING = "b:" + CALLER_BED_NAME;
+    public static final String CALLER_BED_STRING = "my-bed";
+    public static final String OLD_BED_STRING = "playerbed";
     private String playername = "";
     private boolean isValid;
     private Location knownBedLoc;
@@ -47,11 +47,11 @@ public class BedDestination implements MVDestination {
         boolean validFormat = split.length >= 1 && split.length <= 2 && split[0].equals(this.getIdentifier());
 
         OfflinePlayer p = Bukkit.getOfflinePlayer(split[1]);
-        boolean validPlayer = p.getName() != null && !p.getName().equals(CALLER_BED_NAME);
+        boolean validPlayer = p.getName() != null && !(p.getName().equals(OLD_BED_STRING) || p.getName().equals(CALLER_BED_STRING));
 
         if (validFormat && validPlayer) this.playername = p.getName();
 
-        this.isValid = destination.equals(CALLER_BED_STRING) || (validFormat && validPlayer);
+        this.isValid = destination.equals("b:" + CALLER_BED_STRING) || destination.equals("b:" + OLD_BED_STRING) || (validFormat && validPlayer);
 
         return this.isValid;
     }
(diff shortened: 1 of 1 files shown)


nicegamer7:
Thank you for taking the time to look into this issue. I'll give this PR a review sooner or later.

src/main/java/com/onarandombox/MultiverseCore/destination/BedDestination.java
```diff
@@ -46,11 +47,11 @@ public boolean isThisType(JavaPlugin plugin, String destination) {
         boolean validFormat = split.length >= 1 && split.length <= 2 && split[0].equals(this.getIdentifier());
 
         OfflinePlayer p = Bukkit.getOfflinePlayer(split[1]);
-        boolean validPlayer = (p != null);
+        boolean validPlayer = p.getName() != null && !(p.getName().equals(OLD_BED_STRING) || p.getName().equals(CALLER_BED_STRING));
 
         if (validFormat && validPlayer) this.playername = p.getName();
 
-        this.isValid = destination.equals(OLD_BED_STRING) || (validFormat && validPlayer);
+        this.isValid = destination.equals("b:" + CALLER_BED_STRING) || destination.equals("b:" + OLD_BED_STRING) || (validFormat && validPlayer);
```
benwoo1110, line 54: I think we can just decide on one identifier to used as for the caller bed string. I think `playerbed` is ok already, since I googled on mcuuid and there is no such account name.

akrantz01: Fair enough. I had just changed it to ensure that there would never be any conflicts, but its probably pretty unlikely that anyone would register and use the name `playerbed`.

nicegamer7: I kinda disagree. I think we should change it so that there is no possibility for conflict. @dumptruckman any opinion on this? I guess it doesn't really matter... the username `playerbed` is kinda useless.

nicegamer7: Actually, I think that would be a change for MV 5.

nicegamer7 (dismissed):
LGTM 👍.

src/main/java/com/onarandombox/MultiverseCore/destination/BedDestination.java
```diff
@@ -23,7 +23,7 @@
  * A bed-{@link MVDestination}.
  */
 public class BedDestination implements MVDestination {
-    public static final String OLD_BED_STRING = "b:playerbed";
+    public static final String OLD_BED_STRING = "playerbed";
```
benwoo1110, line 26: `OLD_BED_STRING` doesnt sound like a good variable name to me. Something like `OWN_BED_STRING` would be preferred.

benwoo1110 (changes requested):

nicegamer7 (approved):

nicegamer7:
@akrantz01 I just want to ask if you've tested these changes? The changes look good to me, just want to make sure.

akrantz01:
The original changes I tested on 1.16, 1.15, and 1.14, but the most recent ones I haven't been able to test since I'm on the road.

nicegamer7:
Okay, I just tested it and it's all good 👍.

benwoo1110 (approved):

nicegamer7:
Just so you're aware of what's going on @akrantz01, ben and I are waiting on approval from the main dev.

benwoo1110:
Looks good to me, I think I got the greenlight (ish) to merge small bug fixes, so let's merge this 😃

PR #2494, opened 2020-11-27
Fix cannot remove block currency with mvm set command.

### LINKED ISSUE #2484: Can't set price to tp to world in EssentialsX currency- only accepts positive numbers
### Information

* **Server version: This server is running Paper version git-Paper-277 (MC: 1.16.4) (Implementing API version 1.16.4-R0.1-SNAPSHOT)
** 

* **Full output of `/mv version -p`: https://pastebin.com/0gLfDGgE** 

* **Server log: https://pastebin.com/27SrfLPM** 

### Help request
I am trying to set a price to tp to a world (Fair_Lands) in the currency of EssentialsX, which is currently working properly with Jobs and ChestShop.
When I try to do "/mvm set currency -1" I get the following error message, followed by a list of properties I can set; 
"Sorry, You can't set: 'currency'" 
I am not experiencing this issue when I try to set the currency to any other value, and I am op'd. My list of valid world properties is also attached below as well as being contained in the latest.log. When I try to do /mvm set entryfee I get 
"Spawn is the only param with no VALUE.
Type /mv modify ? For help."
I think there's just something about the world properties and how to set this up that isn't making sense to me, else there has to be an issue with the wiki or a bug going on here.
valid world properties https://pastebin.com/dDp0u9ZV
latest.log https://pastebin.com/27SrfLPM
/mvv version p https://pastebin.com/0gLfDGgE
/mvv -b --include-plugin-list https://pastebin.com/UXHsKeer

benwoo1110:
Fixes #2484.

To remove use of block currency we are supposed to set the property to `-1`. However, `-1` is an invalid material type so you cannot run `/mvm set currency -1 [worldname]`. 

Looking into the code, to allow `-1` as expection is probably going to be hard, so I changed it to **setting currency to `0` is now considered as no currency value.** Since, materialID `0` is just AIR, it should be ok. (doubt anyone set air as currency 😆)

Wiki will need to change as well to reflect this.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/configuration/EntryFee.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/configuration/EntryFee.java
@@ -22,6 +22,8 @@ public class EntryFee extends SerializationConfig {
     @Nullable
     private Material currency;
 
+    private final Material DISABLED_MATERIAL = Material.AIR;
+
     public EntryFee() {
         super();
     }
@@ -51,6 +53,9 @@ public class EntryFee extends SerializationConfig {
      */
     @Nullable
     public Material getCurrency() {
+        if (currency == null || currency.equals(DISABLED_MATERIAL)) {
+            return null;
+        }
         return currency;
     }
 
--- a/src/test/java/com/onarandombox/MultiverseCore/TestEntryFeeConversion.java
+++ b/src/test/java/com/onarandombox/MultiverseCore/TestEntryFeeConversion.java
@@ -50,6 +50,10 @@ public class TestEntryFeeConversion {
         WorldProperties props = new WorldProperties(config);
         assertNull(props.entryfee.getCurrency());
 
+        entryFee.put("currency", 0);
+        props = new WorldProperties(config);
+        assertNull(props.entryfee.getCurrency());
+
         entryFee.put("currency", 1);
         props = new WorldProperties(config);
         assertEquals(Material.STONE, props.entryfee.getCurrency());


src/main/java/com/onarandombox/MultiverseCore/configuration/EntryFee.java
```diff
@@ -51,6 +53,9 @@ public double getAmount() {
      */
     @Nullable
     public Material getCurrency() {
+        if (currency == null || currency.equals(DISABLED_MATERIAL)) {
+            return null;
+        }
         return currency;
```
montlikadani, line 59: Simplify these checks into one.
```suggestion
        return currency == Material.AIR ? null : currency;
```

src/main/java/com/onarandombox/MultiverseCore/configuration/EntryFee.java
```diff
@@ -22,6 +22,8 @@
     @Nullable
     private Material currency;
 
+    private final Material DISABLED_MATERIAL = Material.AIR;
```
montlikadani, line 25: Please don't make unecessarry field into the heap.

benwoo1110: I feel this makes make why `Material.AIR` should return null clearer, see what @dumptruckman and @nicegamer7 thinks.

nicegamer7: The amount of memory needed to store an enum is probably negligible. I think since this improves readability, it's fine how it is.

src/main/java/com/onarandombox/MultiverseCore/configuration/EntryFee.java
nicegamer7: Nice suggestion, it's up to ben though.

nicegamer7 (approved):
Good solution IMO 👍.

src/main/java/com/onarandombox/MultiverseCore/configuration/EntryFee.java
benwoo1110: Yea can change that.

src/main/java/com/onarandombox/MultiverseCore/configuration/EntryFee.java
benwoo1110: Yea that's what I thought as well.

benwoo1110:
lets merge this

PR #2498, opened 2020-12-02
/mvlist should list world name alongside alias.

### LINKED ISSUE #1420: [Suggestion] Include option to switch between worldalias and worldname in /mv list
Just a suggestion, it would be cool if you could do /mv list and have either; the world name and world alias displayed OR something in config.yml to decide if it displays world aliases or world names. 

Just thought this might be useful because I am having an issue whereby I use the world aliases in my chat plugin to define what server they're part of (it was easier than trying to code formatting into bungeecord server names) and now /mvlist is pretty useless as it just shows &{colourcode}{Servername}. Example: Donators world becomes &6D for all of the multiverse worlds inside. If I could swap from /mvlist displaying world aliases then that would solve my issue.

benwoo1110:
Fixes #1420

unknown

so the format is: `<worldname> (<alias>) - <environment>`

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
@@ -28,7 +28,6 @@ import org.bukkit.Material;
 import org.bukkit.World;
 import org.bukkit.World.Environment;
 import org.bukkit.WorldType;
-import org.bukkit.command.CommandSender;
 import org.bukkit.configuration.serialization.SerializableAs;
 import org.bukkit.entity.Player;
 import org.bukkit.permissions.Permission;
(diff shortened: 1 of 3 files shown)


src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
```diff
@@ -516,22 +516,39 @@ public World getCBWorld() {
...
+    private boolean hasCustomName() {
+        return props.getAlias() != null
+                && !props.getAlias().isEmpty()
+                && !props.getAlias().equals(name);
+    }
+
+    private boolean hasCustomColor() {
+        return props.getColor() != null
+                && props.getColor().getColor() != null
+                && !props.getColor().getColor().equals(ChatColor.WHITE);
```
montlikadani, line 545: Use null-safe check when you trying to compare enums.
```suggestion
                && props.getColor().getColor() != ChatColor.WHITE;
```

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
```diff
@@ -52,23 +52,38 @@ public ListCommand(MultiverseCore plugin) {
...
+                        .append("[H] ")
+mv lis                        .append(ChatColor.WHITE);
+            }
+
+            builder.append(world.getName());
+
+            if (world.hasAlias()) {
+                builder.append(" (")
+                        .append(world.getColoredWorldString())
+                        .append(")");
```
montlikadani, line 70: Better
```suggestion
                        .append(')');
```

src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
```diff
@@ -516,22 +516,39 @@ public World getCBWorld() {
...
-        nameBuilder.append(props.getAlias()).append(ChatColor.WHITE).toString();
+        }
+        nameBuilder.append(props.getAlias()).append(ChatColor.WHITE);
 
         return nameBuilder.toString();
     }
 
+    private boolean hasCustomName() {
+        return props.getAlias() != null
+                && !props.getAlias().isEmpty()
```
montlikadani, line 538: You don't need `#isEmpty` check here. It will always false if the `getAlias` is not equal to the world name.

benwoo1110: Actually, I made a mistake in the logic here. It should return false if the alias is **either** equal to the world name **or** empty.

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
```diff
@@ -52,23 +52,38 @@ public ListCommand(MultiverseCore plugin) {
...
-                    worldList.add(ChatColor.GRAY + "[H]" + builder.toString());
-                }
-            } else {
-                worldList.add(builder.toString());
+
+            if (world.isHidden() && (p == null || this.plugin.getMVPerms().hasPermission(p, "multiverse.core.modify", true))) {
+                // Prefix hidden worlds with an "[H]"
+                builder.append(ChatColor.GRAY)
+                        .append("[H] ")
+mv lis                        .append(ChatColor.WHITE);
```
benwoo1110, line 62: Whoops, idk what happened here, marking it to fix it when I have the time.

benwoo1110:
Will be done in MV5

PR #2503, opened 2020-12-11
Use getPlayerExact instead of getPlayer

nicegamer7:
`getPlayer` searches for usernames instead of performing a lookup. I tested this by running `/mv check nicegamer world` (my username is `nicegamer7`, not `nicegamer`), and the command ran successfully as if I had entered `nicegamer7`.

This is counter intuitive, so I've changed all occurrences of `getPlayer` to `getPlayerExact` (which is still case insensitive). I did this by searching the codebase with this regex:
```
getPlayer\([^\)]
```
Also, a lookup is obviously faster than searching, so there's that.


benwoo1110 (approved):
👍

PR #2504, opened 2020-12-12
Refactor clone operation with CloneHelper class.

benwoo1110:
This will allow clone to have clearer structure and logging messages. Also want to know how you all think about the way I refactor haha!

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
@@ -47,7 +47,6 @@ import java.util.Arrays;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -96,127 +97,203 @@ public boolean accept(File file, String s) {
...
-        final File newWorldFile = new File(this.plugin.getServer().getWorldContainer(), newName);
-        final List<String> ignoreFiles = new ArrayList<>(Arrays.asList("session.lock", "uid.dat"));
+        /**
+         * Ensure that world with oldName can be clone to a world called newName.
+         *
+         * @return True if parameters are valid, otherwise false.
+         */
+        private boolean validCloneParameters() {
+            String originalOldName = oldName;
+            oldName = isWorldFromConfig(oldName) ? oldName : getWorldByAliasFromConfig(oldName);
```
montlikadani, line 155: Do not try to change the `oldName` variable to the same if `isWorldFromConfig` is true to avoid misunderstands.
```suggestion
           if (!isWorldFromConfig(oldName)) {
               oldName = getWorldByAliasFromConfig(oldName);
           }
```

PR #2505, opened 2020-12-12
Update to all use static logging.

benwoo1110:
Changed all the this.plugin.log to the static Logging class. Should I then remove all the deprecated log methods and LoggablePlugin interface?


nicegamer7 (approved):
LGTM 👍.

PR #2506, opened 2020-12-14
Update mapInitialize to use non deprecated methods.

benwoo1110:
getItemInHand and getDurability are deprecated.


benwoo1110:
Actually thinking of it, is there even a need for this class loll

benwoo1110:
https://hastebin.com/epifurusoc.swift (just incase i need)

PR #2508, opened 2020-12-14
Code cleanup to ChatEvent.

benwoo1110:
Hopefully better readability. We also shouldn't do `isMVWorld` check then `getMVWorld`, when we can getMVWorld and check of null directly.


nicegamer7:
I say we leave this to MV 5.0.0.

The reason is that we want to remove the PlayerChatEvent Listener since its deprecated. When we do that, we can remove a bunch of classes and there'll be less cleanup to do.

benwoo1110:
Yea we can leave it for 5.0.0, but the code for that handling is still the same (unless we changing how chat works) so this will still be applicable then

PR #2510, opened 2020-12-15
Use final fields where needed.

benwoo1110:
probably just my ocd, but good to ensure variables that should not be modified to be set as final.


benwoo1110:

PR #2539, opened 2021-01-23
Add a Config Option to Disable Buscript

nicegamer7:
This PR adds a config option to enable/disable Buscript. This was requested on the Spigot thread, so that users can have a way of removing the warning from the console (the one that pops up when Buscript can't be initialized).


benwoo1110 (approved):
Looks good, but till today I still have no idea how buscript works 😂

benwoo1110:
@nicegamer7, I think you should add the config option to the `mv version` output

nicegamer7:
Will do 👍.

benwoo1110 (approved):
Looks good!

PR #2548, opened 2021-01-27
fix deadly spectator mode...

dc6iq:
Here is the fix to the deadly spectator gamemode problem,

cheers and keep the good work up !


benwoo1110:
Thanks for the PR, appreciate it. However, as stated in the issue, we need to revamp how permission work in mv as a whole, as such we are unable to accept this point in time. If you are interested in helping out I suggest discussing with us on discord https://discord.gg/NZtfKky

PR #2556, opened 2021-02-08
Bump jetbrains annotation to 20.1.0.

benwoo1110:


nicegamer7:
I tried this locally (and also bumped up the intrumentor) and it wouldn't build. Can you try that on your end and tell me if you get the same thing?

benwoo1110:
With java 8 its fine, but I didn't test with java 11 this this PR was made before that issue was reported.

nicegamer7:
I was testing on Java 15. I'll try with 8.

nicegamer7:
It still doesn't build with the updated instrumentor. I guess we can leave this until 5.0.0 unless it fixes #2559.

benwoo1110:
Yep this is something for 5.0.0, just made a PR to solve that issue. #2573

PR #2555, opened 2021-02-08
Update to latest 1.16.5-R0.1-SNAPSHOT dependency.

benwoo1110:
This will break support for 1.14 and below.

* Add minidev json dependency.
* Remove use of travel agent.

I need this for some 5.0.0 PRs.

PR #2564, opened 2021-02-18
Respawn anchor

benwoo1110:
Add respawn anchor support while retaining support for 1.15. 

Will need #2555 to be merged first.


benwoo1110:
Actually this shouldnt need to be in 5.0.0, since I'm using reflection anyways

PR #2568, opened 2021-02-21
Add support for fullname face direction.

benwoo1110:
See a few people is discord typing `north` instead of `n` for facing direction. Figured why not just add support for it, so we can do `w:world:north` on top of the current `w:world:n`.


nicegamer7 (approved):
LGTM 👍.

nicegamer7:
Can you add this to the wiki?

benwoo1110:
sure.

PR #2582, opened 2021-03-04
Update unload message logging.

### LINKED ISSUE #2533: Improve world unload message
### Feature request

**Feature description**
This is a minor request, as it only is about a specific message.

When you either delete or unload a world will MV post the message that the world has been removed from memory twice.

For example, executing `mvdelete example` followed by `mvconfirm` gives back those log entries.
```
[15:12:27] [Server thread/INFO]: [Multiverse-Core] World 'example' was unloaded from memory.
[15:12:27] [Server thread/INFO]: ThreadedAnvilChunkStorage (example): All chunks are saved
[15:12:27] [Server thread/INFO]: [Multiverse-Core] World 'example' was unloaded from memory.
[15:12:27] [Server thread/INFO]: [Multiverse-Core] World 'example' was removed from config.yml
[15:12:27] [Server thread/INFO]: [Multiverse-Core] World 'example' was DELETED.
```
As you can see is the message `World 'example' was unloaded from memory.` posted twice. Once when it starts and once when the world was actually unloaded.

I want to suggest to change the first message to a different one, if doable. Something like `Attempting to unload World 'example' from memory...`

**How the feature is useful**
Mainly to prevent any possible confusion as to why this message pops up twice.
As I said is this a really minor visual issue and barely anyone may notice this, but it would still be good to correct this in some way.

benwoo1110:
Fixes #2533. `World '%s' was unloaded from server.` twice, it should differentiate between unloading from bukkit and multiverse. Just makes it clearly since someone pointed out in #2533.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
@@ -385,14 +385,14 @@ public class WorldManager implements MVWorldManager {
             this.worldsFromTheConfig.get(name).cacheVirtualProperties();
             if (unloadBukkit && this.unloadWorldFromBukkit(name, true)) {
                 this.worlds.remove(name);
-                Logging.info("World '%s' was unloaded from bukkit.", name);
+                Logging.info("World '%s' was unloaded from Bukkit.", name);
                 return true;
             } else if (!unloadBukkit){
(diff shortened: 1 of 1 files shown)


nicegamer7:
Looks good, the lowercase letters were bugging me though lol.

nicegamer7 (approved):

PR #2580, opened 2021-03-04
Improve portal search radius

benwoo1110:
* Adds new ReflectHelper class to help in reflection used in CompatibilityLayer.
* Adds support for 1.15+ method in setting portal search radius, while using Travel Agent for older mc versions.
* Apply portal search radius to entities as well.

Requires #2586 to be merged first.

PR #2586, opened 2021-03-08
Add ReflectHelper class to better facilitate various reflection actions.

benwoo1110:
Various methods to help facility reflection.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/utils/ReflectHelper.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/utils/ReflectHelper.java
@@ -1,0 +1,140 @@
+package com.onarandombox.MultiverseCore.utils;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+/**
+ * Utility class used to help in doing various reflection actions.
+ */
+public class ReflectHelper {
+
+    /**
+     * Try to get the {@link Class} based on its classpath.
+     *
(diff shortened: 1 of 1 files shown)


src/main/java/com/onarandombox/MultiverseCore/utils/ReflectHelper.java
```diff
@@ -0,0 +1,140 @@
...
+     * @return A {@link Class} if found, else null.
+     */
+    @Nullable
+    public static Class<?> getClass(String classPath) {
+        try {
+            return Class.forName(classPath);
+        } catch (ClassNotFoundException e) {
+            return null;
+        }
+    }
```
dumptruckman, line 26: Hopefully we're not using this all the time as it is a relatively costly operation. Class.forName is "slow" and should an exception be thrown that will also slow this down quite a bit.

dumptruckman: We might want to consider a caching strategy here.

benwoo1110: Yea we do store the result, it should only be used on started, then the results are stored.

benwoo1110: You can see #2580

dumptruckman (approved):
ok for now.

nicegamer7 (commented):
Seems good I guess 👍.

PR #2589, opened 2021-03-09
Implement new command flag system.

benwoo1110:
* Add support for potential tab completion.
* Adds support for more descriptive flags, e.g. `/mv create test normal --type flat --structures true --dont-adjust-spawn`. Old flags are retained as well so the equivalent is `/mv create test normal -t flat -a true -n` will still work.

Currently, I only implemented it for the create command, but the idea will basically be the same for all command with flags.


benwoo1110:
Will be done in #2827

PR #2590, opened 2021-03-09
Small WorldManager updates

benwoo1110:
* Some operations do not need to lookup of MVWorld by alias, e.g. when its the bukkit world object or folder name. Since alias is by iteration, it saves performance to not do this check whenever possible.
* Make use of the new WorldNameChecker and added getPotentialWorlds method to WorldManager. More streamlined and useful for tab-complete in ACF.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
@@ -9,7 +9,6 @@ package com.onarandombox.MultiverseCore.commands;
 
 import com.onarandombox.MultiverseCore.MultiverseCore;
 import com.onarandombox.MultiverseCore.api.MVWorldManager;
-import com.onarandombox.MultiverseCore.api.MultiverseWorld;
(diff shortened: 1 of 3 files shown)


src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
```diff
@@ -46,33 +43,17 @@ public ImportCommand(MultiverseCore plugin) {
...
-                    currColor = ChatColor.YELLOW;
-                } else {
-                    currColor = ChatColor.WHITE;
-                }
-            }
+
+        for (String world : potentialWorlds) {
+            worldList.append(currColor).append(world).append(' ');
+            currColor = currColor == ChatColor.WHITE ? ChatColor.YELLOW : ChatColor.WHITE;
         }
```
dumptruckman, line 54: We should probably have a utility method that does this since we do it in several places. Something that takes like vararg colors and a collection of items maybe?

dumptruckman (approved):
Cool.

nicegamer7 (commented):
Looks good 👍.

PR #2567, opened 2021-02-21
Add support to find players based on name, UUID or selectors.

benwoo1110:
Fixes #1961.

This is will also be fixed with ACF, but I think it can be brought forward, and I also want to split some API changes from ACF into smaller PRs.

PR #2591, opened 2021-03-09
Make UUID regex less restrictive.

benwoo1110:
As @nicegamer7 point out in https://github.com/Multiverse/Multiverse-Core/pull/2567#discussion_r589683007, software such as Floodgate have non-standard mc UUIDs, so we should be less strict during UUID regex check.


nicegamer7:
Can you keep the pattern like this: `[0-9a-fA-F]`? Just so the diff is easier to understand.

benwoo1110:
Sure.

nicegamer7 (approved):
Okay, looks good 👍.

PR #2593, opened 2021-03-09
Add method to get registered destination identifiers.

benwoo1110:
Get destination identifiers, useful for operations like tab-complete etc...

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
@@ -21,6 +21,7 @@ import org.bukkit.permissions.PermissionDefault;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
```diff
@@ -164,4 +165,8 @@ public boolean registerDestinationType(Class<? extends MVDestination> c, String
         this.teleportCommand.addAdditonalPermission(other);
         return true;
     }
+    
+    public Collection<String> getRegisteredIdentifiers() {
+        return this.destList.keySet();
+    }
```
dumptruckman, line 171: Easy peasy. But since you were talking about javadocs in discord I think something is missing here...

dumptruckman (approved):

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationFactory.java
nicegamer7: Fixed in #2597.

nicegamer7 (commented):
Looks good.

PR #2602, opened 2021-03-18
Remove ability to unload default worlds.

benwoo1110:
People unloading the default nether causing nether portals to not work on 1.16 is happening far too often. This will prevent that. Also, adds potentially useful API method to get default world names.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
@@ -40,6 +40,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
@@ -59,6 +60,7 @@ public class WorldManager implements MVWorldManager {
     private final WorldPurger worldPurger;
     private final Map<String, MultiverseWorld> worlds;
     private Map<String, WorldProperties> worldsFromTheConfig;
+    private Set<String> defaultWorldNames;
     private FileConfiguration configWorlds = null;
     private Map<String, String> defaultGens;
     private String firstSpawn;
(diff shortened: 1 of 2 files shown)


src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
```diff
@@ -707,16 +714,23 @@ private boolean isMVWorldAlias(final String alias) {
...
+        this.defaultWorldNames = new HashSet<>(myWorlds.size());
         for (World w : myWorlds) {
             String name = w.getName();
             if (!worldsFromTheConfig.containsKey(name)) {
                 String generator = null;
                 if (this.defaultGens.containsKey(name)) {
                     generator = this.defaultGens.get(name);
                 }
-                this.addWorld(name, w.getEnvironment(), String.valueOf(w.getSeed()), w.getWorldType(), w.canGenerateStructures(), generator);
+                if (!this.addWorld(name, w.getEnvironment(), String.valueOf(w.getSeed()), w.getWorldType(), w.canGenerateStructures(), generator)) {
```
montlikadani, line 725: `String.valueOf` -> `Long.toString`

nicegamer7: This makes no difference as far as I am aware. They will both result in the same thing (please correct me if I'm wrong).

nicegamer7:
I didn't review the changes yet, but I'm not sure I like this idea. Maybe we should simply warn that unloading a default world will affect portal usage and ask them to type `/mv confirm`.

src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
benwoo1110: String.valueOf is just using the ToString method. Long.ToString _may_ be different. But isnt part of the change anyways, was String.valueOf.

benwoo1110:
Won’t effecting portal usage be a large enough reason for people to not unload the world?

nicegamer7:
I think so, but I don't think we should disallow it. We should just let them know it will affect portal usage and let them decide.

nicegamer7:
And I just realized, this will prevent people from regenerating default worlds right? I don't think we want that.

benwoo1110:
You can actually regen the default overworld even now, it’s just about the nether and end. It’s weird that they allow but cause breaking issues.

nicegamer7:
You can regenerate the default nether and end though without errors right?

Regarding portals breaking on regening them, I fixed that and sent a PR to Spigot.

benwoo1110:
With the new queue command system merge, I don’t mind having a confirm command for default world unloading / deleting

src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java
montlikadani: `String.valueOf` returns the same when you call `Long#toString`. Yea, obviously does not makes any difference, just making sure that we're not calling two methods.

benwoo1110:
probably not needed

PR #2611, opened 2021-04-06
Make prefixchat config option default to 'false'

Turbotailz:
This may seem like a really negligible change but having this option default to `true` actually causes a lot of pain for support in other plugins that deal with prefixes (LuckPerms and Essentials being the main victims). Users always come in guns blazing, screaming at us because they don't know how to remove the world name from their chat. It's such a simple solution, yet no one thinks that of all plugins, it would be Multiverse that is tainting their beloved chat format.

Please consider updating this default setting. It will save a lot of us trouble in the future 🙏


underscore11code (approved):

benwoo1110 (approved):
😄

PR #2588, opened 2021-03-08
Implement new WorldNameChecker

benwoo1110:
Allow better preliminary checking of world name and folder status. Mainly targeted to ACF and improving command output, but bring it forward to split the journey to ACF a more manageable one.

PR #2619, opened 2021-04-24
General Cleanup

nicegamer7:
This PR just cleans up some small things.

1. Updates the bStats Metrics class to the newest version.
2. Enables `minimizeJar` for `maven-shade-plugin` (reduces JAR size from 487KB to 450KB).
3. Adds to the message that is printed when Buscript fails to load.
4. Updates the VaultAPI repo (the one in our pom is no longer active).
5. Corrects a copy-paste error introduced in #2588.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,10 +34,6 @@
             <id>minebench-repo</id>
             <url>https://repo.minebench.de/</url>
         </repository>
-        <repository>
-            <id>CodeMC</id>
-            <url>https://repo.codemc.org/repository/maven-public</url>
-        </repository>
     </repositories>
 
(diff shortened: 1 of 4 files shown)


benwoo1110 (approved):
Looks good. Assume you tested with the minimisejar, just that make sure no class loading issues

nicegamer7:
I built a JAR, spun up a server, created a world and teleported to it. Do you think that's enough, or should I do more thorough testing? I was wondering this myself.

benwoo1110:
I think it's good enough, my own plugins use it just fine, I think most classes are loaded on startup.

PR #2624, opened 2021-04-30
Abit of 1.12 support

benwoo1110:
This will let mv-core start (and mostly run) without errors. Note it does not mean full 1.12 support for all mv plugins.

What I tested:
- Loading/Creating/Import/Deleting worlds
- Config files save properly
- Currency and pricing works (i.e. block id stuff is all good)
- chat messaging and prefix 

Changes:
- Phantom class
- Materials
- GameRules class

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/utils/MaterialConverter.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/utils/MaterialConverter.java
@@ -32,11 +32,33 @@ public class MaterialConverter {
      */
     @Nullable
     public static Material convertTypeString(@Nullable String value) {
-        IdMappings.Mapping mapping = IdMappings.getById(value != null ? value : "");
-        if (mapping != null) {
-            return Material.matchMaterial(mapping.getFlatteningType());
-        } else {
-            return Material.matchMaterial(value != null ? value : "");
+        String targetValue = value != null ? value : "";
+
+        IdMappings.Mapping idMapping = IdMappings.getById(targetValue);
+        if (idMapping != null) {
+            return getMappedType(idMapping);
         }
+        IdMappings.Mapping flattenMapping = IdMappings.getByFlatteningType(targetValue);
+        if (flattenMapping != null) {
+            return getMappedType(flattenMapping);
+        }
+        IdMappings.Mapping legacyMapping = IdMappings.getByLegacyType(targetValue);
+        if (legacyMapping != null) {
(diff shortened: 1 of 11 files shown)


nicegamer7 (dismissed):
LGTM 👍.

src/main/java/com/onarandombox/MultiverseCore/utils/SimpleWorldPurger.java
```diff
@@ -32,16 +33,8 @@
...
-    private Class<Entity> ambientClass = null;
-
     public SimpleWorldPurger(MultiverseCore plugin) {
         this.plugin = plugin;
-        try {
-            Class entityClass = Class.forName("org.bukkit.entity.Ambient");
-            if (Entity.class.isAssignableFrom(entityClass)) {
-                ambientClass = entityClass;
-            }
-        } catch (ClassNotFoundException ignore) { }
```
nicegamer7, line 48: I'm curious as to why this was ever needed. The Ambient class has been present since at least Bukkit 1.5!

src/main/java/com/onarandombox/MultiverseCore/utils/MaterialConverter.java
```diff
@@ -32,11 +32,33 @@ public static Material convertConfigType(@NotNull ConfigurationSection config, @
      */
     @Nullable
     public static Material convertTypeString(@Nullable String value) {
```
nicegamer7, line 34: I'm pretty confused by the change to this method.

I can see that you've changed it to call `getLegacyType` when necessary, but you also added two more code blocks. What are they for?

src/main/java/com/onarandombox/MultiverseCore/utils/SimpleWorldPurger.java
benwoo1110: mv code was from like 2011, and bukkit 1.5 was like in 2013, ig they just patched it and no one care to change it after that.

src/main/java/com/onarandombox/MultiverseCore/utils/MaterialConverter.java
benwoo1110: Its to support multiple formats, for example, if you want to get a granite material, you can pass in `1:1`, `granite` (on >1.13), or `stone:1` you will get the same material enum.

benwoo1110:
Not planning to support 1.12 anymore

PR #2629, opened 2021-05-16
Improve worldname checks

benwoo1110:
Adds config option to allow unsafe (e.g. / ) to be part of the world name. Those that know what they doing can use it for subdir worlds etc. Also, integrate the new world name restrictions from #2353 to WorldNameChecker to make things cleaner.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/MultiverseCoreConfiguration.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MultiverseCoreConfiguration.java
@@ -1,7 +1,5 @@
 package com.onarandombox.MultiverseCore;
 
-import java.util.Map;
-
 import com.dumptruckman.minecraft.util.Logging;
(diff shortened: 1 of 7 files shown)


benwoo1110:
example output:
image

Test build:
[Multiverse-Core-4.2.3-SNAPSHOT.zip](https://github.com/Multiverse/Multiverse-Core/files/6489321/Multiverse-Core-4.2.3-SNAPSHOT.zip)

nicegamer7:
I don't like the idea of allowing unsafe world names. I liked your idea where we check if the world is in the world container. Would that not work?

benwoo1110:
It will work, just more work haha, but yea its possible. I can try to see if it works out in code.

nicegamer7:
Okay, sounds good.

nicegamer7:
Actually, I think I have a simple solution. A world name must not start with a `/` and must not contain `..`.

Pretty sure that will ensure worlds stay within the server's directory.

PR #2639, opened 2021-06-08
Fixed plugin ignoring settings allow-end and allow-nether in server

### LINKED ISSUE #2634: PlayerPortalEvent firing even if allow-nether is false
* **Server version:** This server is running Paper version git-Paper-734 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)

* **MV Core Version** https://pastebin.com/NQ6w5x2Q

**Description**  
mv core broke the "allow-nether=false" option in server.properties

**Steps to reproduce**  
1. set allow-nether=false in server.properties
2. install mv core (only core not portals)
3. nether portals work again

PapiCapi:
configurations.

Fixes #2634 for end portals. Should also fix for nether ones.


benwoo1110:
I don’t think this should be fixed by mv side, as it looks like an issue with the server software instead.

PR #2650, opened 2021-06-23
add portal intercept option

cs8425:
I want players can:
1. use `/mvtp` to teleport between 2 or more normal world (`multiverse.access.XXX` & `enforceaccess: 'true'`)
2. use home plugin for teleport to any world (`teleportintercept: 'false'`)
3. for each normal world, vanilla nether portal for teleport to nether world, and end portal  for teleport to end world 

yet I can not figure out how to set that (only 1 & 2 works),
so I add a option for this.

permission & config I had used:
* `multiverse.access.world1`
* `multiverse.access.world2`
* `multiverse.teleport.self.w`
* `multiverse.teleport.self.b`
* `enforceaccess: 'true'`
* `teleportintercept: 'false'`
* `portalintercept: 'false'` (this PR)


benwoo1110:
Interesting idea... will need ot discuss with the rest of the devs if it's a viable option to add.

PR #2666, opened 2021-07-06
Ensure https is used for repository.

benwoo1110:
Works for mvinv, should work fine for core as well.


dumptruckman (approved):
Sure, why not.

PR #2672, opened 2021-07-16
Fix Tests on Java 16

nicegamer7:
This PR fixes tests on Java 16. It does so by removing PowerMock, as it turns out we don't need it anymore.

Mockito supports mocking final classes starting with version two, which I believe is what we were using PowerMock for. It's not enabled by default, but all you have to do to enable it is create a text file in the recourses folder with a specific string.

Other than that, I removed unneeded imports, and made the usage of `mock()` and `spy()` consistent (static imports in all the files).

### DIFF
--- a/src/test/java/com/onarandombox/MultiverseCore/utils/TestInstanceCreator.java
+++ b/src/test/java/com/onarandombox/MultiverseCore/utils/TestInstanceCreator.java
@@ -85,12 +85,6 @@ public class TestInstanceCreator {
                     "com.onarandombox.MultiverseCore.MultiverseCore"));
             when(pdf.getAuthors()).thenReturn(new ArrayList<String>());
             core = PowerMockito.spy(new MultiverseCore(mockPluginLoader, pdf, pluginDirectory, new File(pluginDirectory, "testPluginFile")));
-            PowerMockito.doAnswer(new Answer<Void>() {
-                @Override
-                public Void answer(InvocationOnMock invocation) throws Throwable {
-                    return null; // don't run metrics in tests
-                }
-            }).when(core, "setupMetrics");
 
             // Let's let all MV files go to bin/test
             doReturn(pluginDirectory).when(core).getDataFolder();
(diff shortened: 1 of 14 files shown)


src/test/java/com/onarandombox/MultiverseCore/utils/TestInstanceCreator.java
```diff
@@ -69,28 +66,20 @@ public boolean setUp() {
...
-            PluginDescriptionFile pdf = PowerMockito.spy(new PluginDescriptionFile("Multiverse-Core", "2.2-Test",
+            PluginDescriptionFile pdf = spy(new PluginDescriptionFile("Multiverse-Core", "2.2-Test",
                     "com.onarandombox.MultiverseCore.MultiverseCore"));
             when(pdf.getAuthors()).thenReturn(new ArrayList<String>());
-            core = PowerMockito.spy(new MultiverseCore(mockPluginLoader, pdf, pluginDirectory, new File(pluginDirectory, "testPluginFile")));
-            PowerMockito.doAnswer(new Answer<Void>() {
-                @Override
-                public Void answer(InvocationOnMock invocation) throws Throwable {
-                    return null; // don't run metrics in tests
-                }
```
nicegamer7, line 92: I just wanted to point out that this is no longer necessary since we added TestingMode (first line of the `setUp` method).

src/test/java/com/onarandombox/MultiverseCore/utils/TestInstanceCreator.java
```diff
@@ -69,28 +66,20 @@ public boolean setUp() {
...
-            }).when(core, "setupMetrics");
```
nicegamer7, line 93: I just wanted to point out that this is no longer necessary since we added TestingMode (used on the first line of the `setUp` method).

benwoo1110:
hmm the gh action failed with some NPE during test, may want to look into it.

nicegamer7:
Good thing we had CI to catch that lol. The issue was that the tests were running in parallel with the Surefire plugin (which is why I didn't catch this originally, they run normally through IJ). This was causing `TestInstanceCreator.setUp()` to be called more than once before being torn down.

I also cleaned up the test files a bit, removing wildcard imports, deprecated imports, out of order imports... stuff like that. I also let IJ do a code cleanup, it removed some unnecessary casting and added a `final` somewhere lol.

benwoo1110 (approved):
LGTM! Can help do the same for mvinv 😄

nicegamer7:
Yup, will do lol.

PR #2683, opened 2021-08-03
Adds permission node to bypass fly enforcement (#2469)

### LINKED ISSUE #2467: Multiverse Enforce Fly Disabling Bypass Feature, instead of only force gamemode bypass.
### Feature request

I am making this Feature request, because Ben told me within the Discord server that this would be a good idea to post here :)

**Feature description**

Right now, when giving people the permission "mv.bypass.gamemode.<NAME>" you are able to make people bypass Multiverse-Core setting/forcing someone to the gamemode that is setup within the config. However, when Multiverse is setting someone to Survival Mode, also his fly is disabled. The issue I have with a MiniGame is that spectators lose their fly when joining a game that is in a different world. This doesn't happen when giving this bypass permission but that is basically not what I want, I just want to bypass the lose of flight. So not the bypass of gamemode enforcing because people should still be set into survival mode.
So, my question is:
Is it possible to JUST bypass Multiverse removing FLIGHT in a certain world.
So for example a permission: "mv.bypass.flying.<Worldname>
This is different from the allowFlight configurable option by the way!

**How the feature is useful**

As actually already explained before, this can be useful for server owners that run different minigames in different worlds and who struggle with fly mode settings. For players this would mean more games and less struggle with flying mode and an overall more enjoyable gameplay! :D :D 

Thank you so much in advance for your reply! I really enjoy Multiverse and I am using it already for more than 8 years for my own small community server. It never has let me down and it basically works like a charm ❤️ 

Thanks for all the amazing years of development!

(P.S. What I basically would like to have can also be seen here, as is a snippet of code I found within the source code, haha.)
unknown

benwoo1110:
Basically #2469 that fixes #2467 but now moved to mv branch instead of by fork. Current still draft bcu not sure of the breaking behaviour changes.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
@@ -337,7 +337,8 @@ public class MVWorld implements MultiverseWorld {
 
     private Permission permission;
     private Permission exempt;
-    private Permission ignoreperm;
+    private Permission ignoregamemodeperm;
+    private Permission ignoreflyperm;
     private Permission limitbypassperm;
 
     /**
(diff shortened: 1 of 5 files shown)


Waterman1001:
I would like to once again thank you Ben for this PR / Version including the permission node.
It is really handy for me to be able to set this per world and it seems to work like a charm :)

zax71:
This could be ported to MV5, then breaking changes won't matter.

benwoo1110:
I will rethink this as a config option in mv5

PR #2693, opened 2021-08-16
Revamp content display API

benwoo1110:
improve based on @dumptruckman comments

### content display API changes
- Decoupled the parsing of map/list vs displaying of inline/paged into ContentParser and SendHandler. Less complex and easier to manage.
- Decoupled the sending away from ContentDisplay. Now it's in its own SendHandler class. ContentDisplay is only responsible for storing and invoking ContentParser and SendHandler class.
- Remove the messy settings stuff. Now it's just setter methods in the respective handler classes.
- Remove the use of ColorAlternator, it was very retrictive.

### Other additions
- Added filter for gamerules command.


src/main/java/com/onarandombox/MultiverseCore/commands/GamerulesCommand.java
```diff
@@ -61,35 +67,46 @@ public void runCommand(CommandSender sender, List<String> args) {
...
+                .withSendHandler(InlineSendHandler.create()
+                        .withHeader("====[ Gamerules for %s%s%s ]====", ChatColor.AQUA, world.getName(), ChatColor.WHITE)
+                        .withFilter(filter))
+                .send(sender);
     }
 
     private Map<String, Object> getGameRuleMap(World world) {
         Map<String, Object> gameRuleMap = new HashMap<>();
-        for (GameRule<?> rule : GameRule.values()) {
+        Arrays.stream(GameRule.values()).forEach(rule -> {
```
dumptruckman, line 106: Why stream? Generally the overhead isn't worth the cost for small data sets I think.

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
```diff
@@ -43,52 +41,52 @@ public ListCommand(MultiverseCore plugin) {
 
     @Override
     public void runCommand(CommandSender sender, List<String> args) {
-        ContentFilter filter = ContentFilter.DEFAULT;
+        ContentFilter filter = DefaultContentFilter.INSTANCE;
```
dumptruckman, line 44: I think convention might say use a getter method for this.

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
```diff
@@ -43,52 +41,52 @@ public ListCommand(MultiverseCore plugin) {
...
         int page = 1;
 
         // Either page or filter.
         if (args.size() == 1) {
             try {
                 page = Integer.parseInt(args.get(0));
             } catch (NumberFormatException ignore) {
-                filter = new ContentFilter(args.get(0));
+                filter = RegexContentFilter.fromString(args.get(0));
```
dumptruckman, line 52: Nice.

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
```diff
@@ -43,52 +41,52 @@ public ListCommand(MultiverseCore plugin) {
...
+                .withSendHandler(PagedSendHandler.create()
+                        .withHeader("%s====[ Multiverse World List ]====", ChatColor.GOLD)
+                        .withFilter(filter)
+                        .withTargetPage(page))
+                .send(sender);
     }
 
-    private Collection<String> getListContents(@NotNull CommandSender sender) {
-        Player player = (sender instanceof Player) ? (Player) sender : null;
+    private ContentParser newWorldListContentParser() {
```
dumptruckman, line 75: I think that this might be more readable as a separate class. It could just be a private static one in this class maybe?

src/main/java/com/onarandombox/MultiverseCore/display/parsers/ContentParser.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+@FunctionalInterface
+public interface ContentParser {
+    /**
+     * Parse the object to string(s) and add it to the content.
+     *
+     * @param sender    The target which the content will be displayed to.
+     * @param content   The content to display.
+     */
+    void parse(@NotNull CommandSender sender, @NotNull List<String> content);
```
dumptruckman, line 19: I think this might be better if it returns the content list instead of is given one to add to. I think this is just a gut feeling and maybe isn't entirely right but I feel like a list you're given to modify smells a bit more than just returning a list.

src/main/java/com/onarandombox/MultiverseCore/display/ContentDisplay.java
```diff
@@ -1,268 +1,68 @@
...
-         * @param sender The CommandSender to show the display to.
-         */
-        public void show(CommandSender sender) {
-            this.build().show(sender);
-        }
+    public void send(@NotNull CommandSender sender) {
+        Objects.requireNonNull(sendHandler, "No send handler set for content display");
+        List<String> parsedContent = new ArrayList<>();
+        contentParsers.forEach(parser -> parser.parse(sender, parsedContent));
+        sendHandler.send(sender, parsedContent);
```
dumptruckman, line 66: Oh yeah, I think this makes a lot more sense. The builder pattern was a bit awkward for this.

src/main/java/com/onarandombox/MultiverseCore/display/filters/ContentFilter.java
```diff
@@ -0,0 +1,21 @@
...
+     */
+    boolean checkMatch(String value);
+
+    /**
+     * Gets whether content needs to be filtered by this filter.
+     *
+     * @return True if content should be filtered, false otherwise.
+     */
+    boolean needToFilter();
+}
```
dumptruckman, line 21: First glance, this seems like it could easily be a single method, functional interface. If the need to filter is false, then the checkMatch should just return true. In that case, you could also just use the built in [Predicate interface](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html).

src/main/java/com/onarandombox/MultiverseCore/display/handlers/SendHandler.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+@FunctionalInterface
+public interface SendHandler {
+    /**
+     * Sends all the content to the given command sender.
+     *
+     * @param sender    The target which the content will be displayed to.
+     * @param content   The content to display.
+     */
+    void send(@NotNull CommandSender sender, @NotNull List<String> content);
```
dumptruckman, line 19: Nice and simple.

dumptruckman (commented):
Some notes for now. I'll have to look again later. Heading to bed.

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
benwoo1110: Yea use the singleton design pattern

src/main/java/com/onarandombox/MultiverseCore/display/filters/ContentFilter.java
benwoo1110: Yea, then reason I d have the second method, is so that there is no need to loop through the content to filter is there nothing to filter. And when displaying to user, it output whether or not filter was applied to the content.

src/main/java/com/onarandombox/MultiverseCore/commands/GamerulesCommand.java
benwoo1110: yea, probably dont need to change.

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
benwoo1110: yea can change that. Probably use a singleton instance?

benwoo1110: hmm dont really want do use static class so I don't have to pass in the `plugin` instance to the private class, so current just doing something like `private class WorldListContentParser implements ContentParser {` and then `.addContentParser(new WorldListContentParser());`

src/main/java/com/onarandombox/MultiverseCore/display/filters/ContentFilter.java
benwoo1110: Example use cases:
```java
// Only send output of filter string used if avaliable.
if (filter.needToFilter()) {
    sender.sendMessage(String.format("%s[Filter '%s']", ChatColor.GRAY, filter));
}
```

```java
// Only loop to filter if needed, to save performance
if (filter.needToFilter()) {
    return content.stream().filter(filter::checkMatch).collect(Collectors.toList());
}
return content;
```

src/main/java/com/onarandombox/MultiverseCore/commands/ListCommand.java
dumptruckman: What's wrong with passing the plugin instance into the new class's constructor?

benwoo1110: tbh nothing wrong, but then I can't really use a single instance anyways, as not a good idea to store plugin objects in static like this right? Like if a new instance plugin instance is created, it may cause issues?

src/main/java/com/onarandombox/MultiverseCore/display/parsers/ContentParser.java
benwoo1110: The reason why this doesn't just return a list of strings so we only need one common list. Instead of each parsing creating a new list as then adding it to another common list... which is potentially a lot of lists. 

For whether if we delete/modify existing entries from other parsers, either we allow it as a feature, or just instead pass a wrapper around the List<String> that only allow adding of elements.

But all these won't be needed if we decide to only allow one content parser.

benwoo1110:
implemented in #2824

PR #2696, opened 2021-08-18
Fix create command using "test" as the world name.

cyberpwnn:
This is passing the world name to other plugin generators incorrectly and is causing issues. The plugin claims its not a bug with MVC, but it is in this case :P

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
@@ -114,7 +114,7 @@ public class CreateCommand extends MultiverseCommand {
                 // If there was only one arg specified, pad with another empty one.
                 genarray.add("");
             }
-            if (this.worldManager.getChunkGenerator(genarray.get(0), genarray.get(1), "test") == null) {
+            if (this.worldManager.getChunkGenerator(genarray.get(0), genarray.get(1), worldName) == null) {
                 // We have an invalid generator.
                 sender.sendMessage("Invalid generator! '" + generator + "'. " + ChatColor.RED + "Aborting world creation.");
                 return;


benwoo1110:
What exact issues its causing? I'd like a better explanation, tho I do agree to check if the generator is functional with the actual target worldname is better regardless.

cyberpwnn:
Custom world generator needs to know what folder the world will go into, it uses the world name to do this

```
@Override
public ChunkGenerator getDefaultWorldGenerator(String worldName, String id) {
    Folder worldFolder = new File(worldName);
    // install needed data
    
    return new CustomPluginWorldGen();
}
```

But multiverse core is passing the param "test" for worldName instead of the real name. Reason I need the real name if your curious is because the generator can support headless generation (directly writes to MCA files skipping bukkit api). Though, MVC should follow the contract as the plugins do for the Plugin interface.

benwoo1110:
yea but that is just a **test check**, that line of code is not used for creating the actual world. The real-world name to generate the actual world will be done later with the WorldManager#addWorld.

cyberpwnn:
Correct, The generator plugin is only aware of the real name when the first chunk is told to generate because the world is passed in, however for this generator that's too late basically. All i was trying to do was get MVC to pass in the actual world name from the arguments, I'm not aware of any processing on the name of the world or if there were any steps before hand that altered the name in any way. So if you see something, or know a better way to simply pass in the world name instead of "test" that would be helpful.

benwoo1110:
I'm still trying to figure out what is not working. MV uses this https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/WorldCreator.html#generator(java.lang.String) to the specific generator for the world to be created. I still don't understand which part is being "too late" and which part of your generate different greatly from the rest since the current setup has being working with almost all bukkit generators for years.

cyberpwnn:
Also, the getDefaultWorldGenerator() method by default returns null, so there's your test right there. If you are testing for a generator without exceptions being thrown on creation, give it correct information to truly test if it will generate when you add the world. Otherwise, a world generator in this situation has to support MVC's edge cases.

cyberpwnn:
It is failing at your test because you are giving it incorrect information so it never gets to the generation part. By giving it correct information, it will pass the test and create the real generator.

benwoo1110:
Even if I pass in the actual world name, at that point in time, the world folder and its contents have not been created anyways. So it seems you only return a valid chunk generator if the world folder exists...?

cyberpwnn:
The only way I can think of solving this on my end is basically trying to catch MVC throwing a test request and not actually create a real generator. It kind of makes the test pointless but it would work.

```java
@Override
public ChunkGenerator getDefaultWorldGenerator(String worldName, String id) {
   if(worldName.equals("test")){
       try{throw new RuntimeException();}catch(e){
            if(e.getStackTraceElements()[0].toString().contains("MultiverseCore")
            {
                return new DummyGenerator();
            }
        }
   }

    return new RealGenerator(worldName, id);
}
```

But again, this might be a little ridiculous. It's called `getDefaultWorldGenerator`

cyberpwnn:
The generator initializes (for bukkit worlds) after the generator data is installed (generator is designed in configs basically) if the configs aren't in the world, it can't generate. So i'm having to do back-flips to re-install twice just for multiverse core.

benwoo1110:
So you have tested, by changing that one line, everything works for you?

cyberpwnn:
The test goes through, works fine.

cyberpwnn:
Though I could just keep the patch I have on my end, up to you.

benwoo1110:
This `this.worldManager.getChunkGenerator(genarray.get(0), genarray.get(1), "test")` should have complete 0 linkage with the actual world created.

PR #2708, opened 2021-09-22
Remove portal deprecated event usage

gplushi:
I fixed the deprecated event by modifying some files referencing the deprecated event and deleting files that 100% depend on removed classes. I also updated the bukkit api to 1.15. Tested and can teleport between worlds on a 1.16.5 Tuinity server.


benwoo1110:
Will this cause 1.13 and 1.14 support to be dropped?

gplushi:
Lets see

gplushi:
org.bukkit.plugin.UnknownDependencyException: Unsupported API version 1.15 seems not, could maybe changing the API version to 1.13 when it was built on 1.15 would work...

gplushi:
Changing API version in plugin.yml to 1.13 while staying on 1.15.2 in POM worked on a Spigot 1.13 server with just multiverse core.

gplushi:
> Will this cause 1.13 and 1.14 support to be dropped?

Commit 5d9ea1e allows them to work

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPortalListener.java
```diff
@@ -38,13 +37,13 @@ public MVPortalListener(MultiverseCore core) {
...
     @EventHandler
-    public void entityPortalCreate(EntityCreatePortalEvent event) {
+    public void entityPortalCreate(PortalCreateEvent event) {
         if (event.isCancelled() || event.getBlocks().size() == 0) {
             return;
         }
         MultiverseWorld world = this.plugin.getMVWorldManager().getMVWorld(event.getEntity().getWorld());
         // We have to do it like this due to a bug in 1.1-R3
-        if (world != null && !world.getAllowedPortals().isPortalAllowed(event.getPortalType())) {
+        if (world != null && !world.getAllowedPortals().isPortalAllowed(PortalType.CUSTOM)) {
```
nicegamer7, line 46: This seems incorrect to me.

nicegamer7 (commented):
I noticed one change that seemed incorrect to me. Additionally, in order to properly support 1.13, we'd have to come up with another solution to set the search radius. Removing the travel agent will break that functionality on 1.13.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPortalListener.java
gplushi: I did that because I was getting an error there.

gplushi:
yeah now my friend cant change worlds i dont know what im doing

nicegamer7:
No worries, thanks for trying.

PR #2717, opened 2021-10-01
Fix velocity not applying.

benwoo1110:
Minecraft seems to rate limit movement which will prevent cannon destination from pushing the player forward. Made a simple solution to just delay it by one tick. Also made some small efficiency improvements.


benwoo1110:
Yea, the guy (or girl) that report the bug to me also tested that the issue is fixed.

PR #2747, opened 2021-12-30
use Collections' factory method instead of static field

KisaragiEffective:
this resolves some unchecked generics warning


KisaragiEffective:
bump

benwoo1110 (approved):
Looks good

PR #2799, opened 2022-08-14
Improve world regeneration command.

### LINKED ISSUE #2783: Keep selected files when regening world
### Describe the feature

PaperMC now places a `paper-world.yml` file inside each world's directory:

```
world_nether/
├── DIM-1/
├── level.dat
├── paper-world.yml <-- new
├── session.lock
└── uid.dat
```
Announcement: https://forums.papermc.io/threads/paper-1-19.344/

It would be good to make `/mv regen` not lose that file.

### How is the feature useful to all Multiverse users?

As people upgrade to 1.19 and start using `/mv regen` they will most likely want to keep their world customizations.

### Agreements

- [X] I have searched for and ensured there isn't already an open issue regarding this.

TraceLosu:
Fixes #2783, also fixes an unwanted behaviour where the spawn on regenerated worlds stays the same when providing a new seed.

When fixing #2783, I don't know if I'm allowed to modify the classes in the API package. If this shouldn't happen, we could always make the `deleteWorldFolder` boolean to keep the paper config instead of deleting all files in the directory.

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/api/MVWorldManager.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/api/MVWorldManager.java
@@ -7,6 +7,7 @@
 
 package com.onarandombox.MultiverseCore.api;
 
+import com.onarandombox.MultiverseCore.enums.KeepWorld;
 import com.onarandombox.MultiverseCore.utils.PurgeWorlds;
 import com.onarandombox.MultiverseCore.utils.SimpleWorldPurger;
 import org.bukkit.World;
@@ -108,11 +109,10 @@ public interface MVWorldManager {
      * @param name The name of the world to remove
      * @param removeFromConfig If true(default), we'll remove the entries from the
      *                         config. If false, they'll stay and the world may come back.
-     * @param deleteWorldFolder If true the world folder will be completely deleted. If false
-     *                          only the contents of the world folder will be deleted
+     * @param keepContents The files you want to keep. (Paper configuration, World directory, or nothing)
      * @return True if success, false if failure.
      */
-    boolean deleteWorld(String name, boolean removeFromConfig, boolean deleteWorldFolder);
+    boolean deleteWorld(String name, boolean removeFromConfig, KeepWorld keepContents);
 
     /**
      * Unload a world from Multiverse.
--- a/src/main/java/com/onarandombox/MultiverseCore/enums/KeepWorld.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/enums/KeepWorld.java
@@ -1,0 +1,19 @@
+package com.onarandombox.MultiverseCore.enums;
+
+/**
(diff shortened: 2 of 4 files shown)


src/main/java/com/onarandombox/MultiverseCore/enums/KeepWorld.java
```diff
@@ -0,0 +1,19 @@
+package com.onarandombox.MultiverseCore.enums;
+
+/**
+ * Custom enum for method argument in deciding what to keep when deleting world files.
+ */
+public enum KeepWorld {
```
benwoo1110, line 6: hmm, maybe it is better to have the enum for "what to delete", as compared to "what to keep". Perhaps a WorldDeleteMode enum with ALL, WORLD_AND_CONFIG, WORLD.

src/main/java/com/onarandombox/MultiverseCore/api/MVWorldManager.java
```diff
@@ -108,11 +109,10 @@ boolean addWorld(String name, Environment env, String seedString, WorldType type
      * @param name The name of the world to remove
      * @param removeFromConfig If true(default), we'll remove the entries from the
      *                         config. If false, they'll stay and the world may come back.
-     * @param deleteWorldFolder If true the world folder will be completely deleted. If false
-     *                          only the contents of the world folder will be deleted
+     * @param keepContents The files you want to keep. (Paper configuration, World directory, or nothing)
      * @return True if success, false if failure.
      */
-    boolean deleteWorld(String name, boolean removeFromConfig, boolean deleteWorldFolder);
+    boolean deleteWorld(String name, boolean removeFromConfig, KeepWorld keepContents);
```
benwoo1110, line 115: Yep definitely should not change the existing interface signature. I suggest marking this as `@deprecated` and creating an overload with the new enum.

benwoo1110 (changes requested):
Haven't tested yet, but here are some pointers based on looking at the code changes.

@nicegamer7 @zax71, if you all happen to have time, can help to try it out.

FakEEE14:
hi 
this is very nice thing, why it's not approved yet ?

zax71:
We are trying to get development back on track internally. As soon as we can review the code as a team this will get merged :)

FakEEE14:
thanks sir

pjindras:
Any update in this case? I would greatly appreciate this improvement :D

ccuser44 (approved):

ken-kentan:
@benwoo1110 Would you mind to review this PR again? We are really waiting this improvement for long..

zax71:
IIRC this feature is in MV5, if not then this should be ported to that branch as no development is occurring on MV4 any more

ken-kentan:
> no development is occurring on MV4 any more

Got it, then I will fork and patch by myself. Thank you for the info!

PR #2805, opened 2022-08-24
Add ignoregamemoderestriction config (adds option to ignore gamemode restriction globally)

soshimee:


benwoo1110:
Hi, thanks for the PR, but this will be implemented in MV5, you can track the progress here https://github.com/Multiverse/Multiverse-Core/pull/2824

PR #2825, opened 2023-02-02
Update actions

Chris6ix:
Update Actions to get rid of deprecation warnings
Also added links to their release pages as a comment to easily find their latest version in the future
image

Builds successfully without the warnings. [My Test Build](https://github.com/Chris6ix/Multiverse-Core/actions/runs/4077431786).


benwoo1110 (approved):
Looks like the build passes with the dependency bump. Thanks!

Chris6ix:
No problem :)

PR #2826, opened 2023-02-02
Switch from Maven to Gradle.

dumptruckman:


build.gradle
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    api('me.main__.util:SerializationConfig:1.7') {
+        exclude group: 'org.bukkit', module: 'bukkit'
+    }
+    api('com.pneumaticraft.commandhandler:CommandHandler:11') {
+        exclude group: 'org.bukkit', module: 'bukkit'
+        exclude group: 'junit', module: 'junit'
+    }
+    api 'com.dumptruckman.minecraft:buscript:2.0-SNAPSHOT'
```
zax71, line 60: May want to remove this as Buscript has been removed now in the MV5 branch

zax71 (commented):
Just one thing

build.gradle
benwoo1110: This is targeting main, so it should be kept. I will change when I merge MV5

.github/workflows/build.yml
```diff
@@ -13,10 +13,17 @@ jobs:
...
-    - name: Build with Maven
-      run: mvn -B package --file pom.xml
+    - name: Build with Gradle
+      run: gradlew build
+
+    - name: Upload Artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: Multiverse-Core-SNAPSHOT
+        path: build/libs/Multiverse-Core-4.3.2-SNAPSHOT.jar
```
benwoo1110, line 30: The version `4.3.2-SNAPSHOT` probably shouldn't be hard coded

zax71: oops, good spot

PR #2827, opened 2023-02-03
feat: Implement command flag system

benwoo1110:
image

Test Build:
[Multiverse-Core-5.0.0-SNAPSHOT.zip](https://github.com/Multiverse/Multiverse-Core/files/10609017/Multiverse-Core-5.0.0-SNAPSHOT.zip)


src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -757,7 +757,7 @@ private void registerCommands() {
         this.commandHandler.registerCommand(new VersionCommand(this));
         this.commandHandler.registerCommand(new ListCommand(this));
         this.commandHandler.registerCommand(new InfoCommand(this));
-        this.commandHandler.registerCommand(new CreateCommand(this));
+        // this.commandHandler.registerCommand(new CreateCommand(this));
```
dumptruckman, line 760: ?

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Subcommand("create")
+    @CommandPermission("multiverse.core.create")
+    @CommandCompletion("WORLDNAME  @flags:groupName=mvcreate")
+    @Syntax("<name> <env> -s [seed] -g [generator[:id]] -t [worldtype] [-n] -a [true|false]")
+    @Description("") //TODO
+    public void onCreateCommand(CommandIssuer issuer,
+
+                                @Flags("trim")
```
dumptruckman, line 87: What is this?

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagGroup.java
```diff
@@ -0,0 +1,132 @@
...
+ * A group of flags.
+ */
+public class FlagGroup {
+    /**
+     * A builder for a flag group.
+     *
+     * @param name The name of the group.
+     * @return The builder.
+     */
+    public static @NotNull Builder builder(@NotNull String name) {
```
dumptruckman, line 21: Is the name always a command name? Seems like that is the intent. Maybe make that more clear here?

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagGroup.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Get the keys and flags contained within in this group.
+     *
+     * @return The keys and flags in this group.
+     */
+    public @NotNull Map<String, MVFlag>getKeysFlagMap() {
+        return keysFlagMap;
+    }
```
dumptruckman, line 76: These expose mutable collections. Maybe consider if there's a way to just provide methods to do the needed things with the collections rather than returning the collections themselves?

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -0,0 +1,110 @@
...
+import org.bukkit.plugin.Plugin;
+import org.jetbrains.annotations.NotNull;
+
+@CommandAlias("mv")
+public class CreateCommand extends MultiverseCommand {
+
+    public CreateCommand(@NotNull MultiverseCore plugin) {
+        super(plugin);
+
+        this.plugin.getCommandManager().getFlagsManager().registerFlagGroup(FlagGroup.builder("mvcreate")
```
dumptruckman, line 37: Maybe `MultiverseCommand` needs a protected shortcut method for registering a flag group.

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagsManager.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Set;
+
+import co.aikar.commands.InvalidCommandArgument;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Manages all the flag groups and parsing.
+ */
+public class FlagsManager {
```
dumptruckman, line 18: Kind of wondering if we should prefix all these flag things with `Command`... `CommandFlagManager` etc.

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagGroup.java
benwoo1110: yea there is, the name is just a unique identifier for getting a specific flaggroup, see the "mvcreate"

https://github.com/Multiverse/Multiverse-Core/blob/8b28cebedf4e7a46eb999af84ba54f9a5fd38f3b/src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java#L101

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagsManager.java
benwoo1110: yea actually that sounds better

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
benwoo1110: yea, i prob adding wordmanager shortcut also, was just "makeitwork" first

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagsManager.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Parses the flags.
+     *
+     * @param groupName The target flag group name.
+     * @param flags The flags to parse.
+     * @return The parsed flags.
+     *
+     * @throws InvalidCommandArgument If the flags are invalid.
+     */
+    public @NotNull ParsedFlags parse(@Nullable String groupName, @NotNull String[] flags) {
```
dumptruckman, line 105: Any way to break this method down and make it smaller? Lot of complexity in this thing. Maybe a separate `FlagParser` is needed to hold the operating state.

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/MVFlag.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents a flag.
+ */
+public class MVFlag {
```
dumptruckman, line 12: Kind of leaning towards doing away with `MV` prefixes on class names.

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
benwoo1110: Ahh something i copied from my old ACF branch, it will be passed to the Context resolver to input sanitize the worldname. (which will be implemented soon)

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/MVFlag.java
benwoo1110: Maybe `CmdFlag` or `CommandFlag`?

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagsManager.java
benwoo1110: yea the logic is quite confusing, but works :laughing:

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
dumptruckman: If `"trim"` is suppose to be a particular option we should probably make it not a string.

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagGroup.java
dumptruckman: Right... But if this is always only ever used for commands could we make it not an arbitrary string?

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/MVFlag.java
dumptruckman: Definitely `CommandFlag` rather than `CmdFlag`. Abbreviations in names should be avoided in general.

src/main/java/com/onarandombox/MultiverseCore/commandtools/flags/FlagGroup.java
benwoo1110: image

mainly bcu to target a specific flaggroup for autocomplete, it is parsed as a string

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
benwoo1110: yea im removing it first, will work on it seperately

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -34,12 +33,12 @@ public class CreateCommand extends MultiverseCommand {
     public CreateCommand(@NotNull MultiverseCore plugin) {
         super(plugin);
 
-        this.plugin.getCommandManager().getFlagsManager().registerFlagGroup(FlagGroup.builder("mvcreate")
-                .add(MVValueFlag.builder("--seed", String.class)
+        this.flagsManager.registerFlagGroup(CommandFlagGroup.builder("mvcreate")
```
dumptruckman, line 36: Why not just `registerFlagGroup`? Also, can we drop the `this`?

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -98,7 +96,7 @@ public void onCreateCommand(CommandIssuer issuer,
                                 @Description("") //TODO
                                 String[] flags
     ) {
-        ParsedFlags parsedFlags = this.plugin.getCommandManager().getFlagsManager().parse("mvcreate", flags);
+        ParsedCommandFlags parsedFlags = this.plugin.getCommandManager().getFlagsManager().parse("mvcreate", flags);
```
dumptruckman, line 99: If `MultiverseCommand` has a `registerFlagGroup` method, maybe it could store the group name and then just have a `parseFlags(flags)` call here.

benwoo1110: the issue is with auto complete, there isnt a way to get the flaggroup from ACF's autocomplete handler, currently this is what is needed. There also isnt a way to get the command class from the CommandCompletionContext (unless we do reflection which is what i wanted to avoid).

https://github.com/Multiverse/Multiverse-Core/blob/85f92bbbc5c2f09362cffcfc63ae6732c7762695/src/main/java/com/onarandombox/MultiverseCore/commandtools/MVCommandCompletions.java#L19-L23

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
benwoo1110: hmm yea

uhh its just a habit that i put this for all class field, so i didnt confuse myself when i was new to java

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
benwoo1110: ohhh wait, nvm i get what you mean, just a shortcut method in multiverse commands

src/main/java/com/onarandombox/MultiverseCore/commands/MultiverseCommand.java
```diff
@@ -15,9 +17,20 @@ public class MultiverseCommand extends BaseCommand {
...
+    private String flagGroupName;
+
     protected MultiverseCommand(@NotNull MultiverseCore plugin) {
         this.plugin = plugin;
         this.worldManager = plugin.getMVWorldManager();
         this.flagsManager = plugin.getCommandManager().getFlagsManager();
     }
+
+    protected void registerFlagGroup(@NotNull CommandFlagGroup flagGroup) {
+        flagsManager.registerFlagGroup(flagGroup);
```
dumptruckman, line 29: Awesome. Only other thing I think is maybe preventing this from being called again.

benwoo1110: Make sense, shoulda thought of that

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                        .completion(() -> {
+                            List<String> types = new ArrayList<>();
+                            for (WorldType type : WorldType.values()) {
+                                types.add(type.name().toLowerCase());
+                            }
+                            return types;
+                        })
+                        .build())
+                .add(CommandFlag.builder("--adjust-spawn")
+                        .addAlias("-n")
```
zax71, line 71: Seams this should be `-a`

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                        .build())
+                .add(CommandFlag.builder("--no-structures")
+                        .addAlias("-a")
```
zax71, line 74: Seams this should be `-n`

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
```diff
@@ -0,0 +1,108 @@
...
+        registerFlagGroup(CommandFlagGroup.builder("mvcreate")
+                .add(CommandValueFlag.builder("--seed", String.class)
+                        .addAlias("-s")
+                        .completion(() -> Collections.singleton(String.valueOf(new Random().nextLong())))
+                        .build())
+                .add(CommandValueFlag.builder("--generator", String.class)
+                        .addAlias("-g")
+                        .completion(() -> Arrays.stream(Bukkit.getServer().getPluginManager().getPlugins())
+                                .filter(Plugin::isEnabled)
+                                .filter(genplugin -> this.plugin.getUnsafeCallWrapper().wrap(
```
zax71, line 45: I'm not good enough to see what is wrong with the `.wrap()` but IntelliJ doesn't like it for some reason. Though it still compiles, so it could be a non-issue.

benwoo1110: ehh yea, really want to change the UnsafeCallWrapper but not part of this pr

src/main/java/com/onarandombox/MultiverseCore/commands/CreateCommand.java
benwoo1110: Well not my fault, i need legacy support also ;-;; Clearly the previous mv devs use weird flags, but with autocomplete should be fine

```java
        boolean allowStructures = true;
        String structureString = CommandHandler.getFlag("-a", args);
        if (structureString != null) {
            allowStructures = Boolean.parseBoolean(structureString);
        }
        String typeString = CommandHandler.getFlag("-t", args);
        boolean useSpawnAdjust = true;
        for (String s : args) {
            if (s.equalsIgnoreCase("-n")) {
                useSpawnAdjust = false;
            }
        }
```

PR #2833, opened 2023-02-06
Destination revamp

benwoo1110:
## Main features
- Support for tab completion
- Support for finer permission such as per world teleport permission

## Tasks
- [x] Implement finer permissions
- [x] Implement Teleport command with ACF
- [x] Implement Check command with ACF
- [x] Implement ACF autocomplete
- [x] Convert all destinations to new API
- [x] Remove old destination API


src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -362,14 +361,13 @@ private void initializeBuscript() {
...
     private void initializeDestinationFactory() {
-        this.destFactory = new DestinationFactory(this);
-        this.destFactory.registerDestinationType(WorldDestination.class, "");
-        this.destFactory.registerDestinationType(WorldDestination.class, "w");
-        this.destFactory.registerDestinationType(ExactDestination.class, "e");
-        this.destFactory.registerDestinationType(PlayerDestination.class, "pl");
-        this.destFactory.registerDestinationType(CannonDestination.class, "ca");
-        this.destFactory.registerDestinationType(BedDestination.class, "b");
-        this.destFactory.registerDestinationType(AnchorDestination.class, "a");
+        this.destinationsManager = new DestinationsManager(this);
```
dumptruckman, line 364: Nitpick: `XManager` is a super generic name without a lot of actual meaning. I would love to avoid names like this in general and try to shoot for names that are more descriptive of their actual purpose.

dumptruckman: Further rationale: Generic names very easily lead to seeming like a catch all for any behavior related to `X` (destinations in this case). Overtime they fill up with functionality that is not related to their original purpose.

src/main/java/com/onarandombox/MultiverseCore/api/Destination.java
```diff
@@ -0,0 +1,19 @@
...
+import co.aikar.commands.BukkitCommandIssuer;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public interface Destination<T extends DestinationInstance> {
+    @NotNull String getIdentifier();
+
+    @Nullable T getDestinationInstance(String destParams);
+
+    @NotNull Collection<String> suggestDestinations(@NotNull BukkitCommandIssuer issuer, @Nullable String destParams);
```
dumptruckman, line 14: Nitpick: Let's not abbreviate `destination` in these parameter names.

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
benwoo1110: hmm good point, `DestinationProvider` prob better describe what the class does i think

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationsManager.java
```diff
@@ -0,0 +1,139 @@
...
+        pluginManager.addPermission(new Permission(PERMISSION_PREFIX + "other." + destination.getIdentifier()));
+    }
+
+    public Collection<String> suggestDestinations(@NotNull BukkitCommandIssuer issuer, @Nullable String deststring) {
+        return destinationMap.values().stream()
+                .filter(destination -> issuer.hasPermission(PERMISSION_PREFIX + "self." + destination.getIdentifier())
+                        || issuer.hasPermission(PERMISSION_PREFIX + "other." + destination.getIdentifier()))
+                .map(destination -> destination.suggestDestinations(issuer, deststring).stream()
+                        .map(s -> destination.getIdentifier() + SEPARATOR + s)
+                        .collect(Collectors.toList()))
```
dumptruckman, line 50: Why is there a stream within a stream?

benwoo1110:
Looks good enough, merging so i can move on to the next change :sweat_smile:

PR #2836, opened 2023-02-10
feat!: Revamp permission handling and its relevant checks

benwoo1110:
Streamline and refactor the permission and access checking.

The main new api classes are:
- `OperationResult`
- `OperationResultChain`
- `PlayerActionChecker`
- `PermissionsRegistrar`
- `PermissionsChecker`

These changes also happen to fix multiple edge-case bugs:
- Double charging the going through portal
- Charging even when player limit has reached

## THINGS TO TEST
- Teleporting between same and different world destinations
- World access (`multiverse.access.<worldname>`)
- World entry fee (`multiverse.exempt.<worldname>`)
- World player limit (`mv.bypass.playerlimit.<worldname>`)
- World blacklist
- Game mode bypass (`mv.bypass.gamemode.<worldname>`)
- Teleport intercept configuration (whether mv should do the above checks if another plugin teleport players)

**TEST BUILD:**
[Multiverse-Core-5.0.0-SNAPSHOT.zip](https://github.com/Multiverse/Multiverse-Core/files/10705355/Multiverse-Core-5.0.0-SNAPSHOT.zip)


src/main/java/com/onarandombox/MultiverseCore/anchor/AnchorManager.java
```diff
@@ -140,18 +140,15 @@ public Set<String> getAnchors(Player p) {
...
-                Logging.finer(String.format("Not adding anchor %s to the list, user %s doesn't have the %s " +
-                        "permission and 'enforceaccess' is enabled!",
-                        anchor, p.getName(), worldPerm));
+            MVWorld mvWorld = this.plugin.getMVWorldManager().getMVWorld(ancLoc.getWorld());
+            if (mvWorld != null) {
+                if (!this.plugin.getPlayerActionChecker().hasAccessToWorld(p, mvWorld).isSuccessful()) {
+                    Logging.finer("Player '%s' does not have access to anchor '%s' in world '%s'.",
+                            p.getName(), anchor, mvWorld.getName());
+                    continue;
+                }
```
dumptruckman, line 149: It would be nice to extract this out to a new method to enhance readability.

src/main/java/com/onarandombox/MultiverseCore/api/MVDestination.java
```diff
@@ -1,142 +0,0 @@
...
-import org.bukkit.entity.Entity;
-import org.bukkit.plugin.java.JavaPlugin;
-import org.bukkit.util.Vector;
-
-/**
- * A destination API for Multiverse
- * Any plugin can add these to MV and when they are, any action that uses them (portals, MVTP, etc.) can use them!
- */
-@Deprecated
-public interface MVDestination {
```
dumptruckman, line 20: Why is this being removed in this PR?

benwoo1110: I wanted to remove it in the destination revamp last time, but it was reference in old permission classes that was revamped in this PR. So now it can be removed.

src/main/java/com/onarandombox/MultiverseCore/anchor/AnchorManager.java
benwoo1110: Yea, but mainly bcu I want to revamp anchors anyways. I want to create an anchor object which contains the anchor name and location.

src/main/java/com/onarandombox/MultiverseCore/api/MVWorld.java
```diff
@@ -200,34 +199,6 @@ public interface MVWorld {
...
-     * @return The permission required to be exempt from charges to/from this world.
-     */
-    Permission getAccessPermission();
-
-    /**
-     * Gets the permission required to be exempt when entering.
-     *
-     * @return The permission required to be exempt when entering.
-     */
-    Permission getExemptPermission();
```
dumptruckman, line 228: Good, these should not be concerns of a World.

src/main/java/com/onarandombox/MultiverseCore/api/action/ActionResponse.java
```diff
@@ -0,0 +1,107 @@
...
+        this(false);
+    }
+
+    /**
+     * @param continueIfFail If true, the response will continue to add results even if one of the results is not successful.
+     */
+    public ActionResponse(boolean continueIfFail) {
+        this.results = new HashSet<>();
+        this.continueIfFail = continueIfFail;
+    }
```
dumptruckman, line 42: Does this have to be public? We should prefer private constructors in general.

src/main/java/com/onarandombox/MultiverseCore/api/action/ActionResponse.java
```diff
@@ -0,0 +1,107 @@
...
+        return addResult(resultSupplier.get());
+    }
+
+    /**
+     * Add a result to the response.
+     *
+     * @param result    The result.
+     * @return self
+     */
+    public @NotNull ActionResponse addResult(ActionResult result) {
```
dumptruckman, line 63: Why is both this method and the `then` method public?

benwoo1110: Hmmmm, actually maybe actionresponse should have a builder?

Then is used, so that if continueIfFail is false, the checks won't be ran. But I still want to keep the option if we need to just add the result directly.

src/main/java/com/onarandombox/MultiverseCore/api/action/ActionResponse.java
benwoo1110: Yea protected makes sense

src/main/java/com/onarandombox/MultiverseCore/commands/TeleportCommand.java
```diff
@@ -35,6 +35,10 @@ public void onTeleportCommand(BukkitCommandIssuer issuer,
     ) {
         // TODO Add warning if teleporting too many players at once.
         for (Player player : players) {
+            if (!this.plugin.getPlayerActionChecker().canUseDestinationToTeleport(issuer.getIssuer(), player, destination).isSuccessful()) {
+                issuer.sendMessage("You don't have permission to use this destination.");
+                continue;
+            }
```
dumptruckman, line 41: Should this not be checking `canGoToDestination` like the check command?

src/main/java/com/onarandombox/MultiverseCore/api/action/ActionResponse.java
```diff
@@ -0,0 +1,107 @@
...
+    public boolean isSuccessful() {
+        return isSuccessful;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean hasResult(@Nullable ActionResult result) {
+        return results.contains(result);
```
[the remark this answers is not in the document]
benwoo1110: Hm, but the java Object itself should have default implement of it, no?

src/main/java/com/onarandombox/MultiverseCore/commands/TeleportCommand.java
benwoo1110: I'm moving it to ACF command condition handling, but mainly bcu not the main scope of this PR, so I put it here for testing, just to make sure the method logic works.

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationsProvider.java
```diff
@@ -45,13 +42,7 @@ public DestinationsProvider(@NotNull MultiverseCore plugin) {
...
     public void registerDestination(@NotNull Destination<?> destination) {
         this.destinationMap.put(destination.getIdentifier(), destination);
-        this.registerDestinationPerms(destination);
-    }
-
-    private void registerDestinationPerms(@NotNull Destination<?> destination) {
-        PluginManager pluginManager = this.plugin.getServer().getPluginManager();
-        pluginManager.addPermission(new Permission(PERMISSION_PREFIX + "self." + destination.getIdentifier()));
-        pluginManager.addPermission(new Permission(PERMISSION_PREFIX + "other." + destination.getIdentifier()));
+        this.plugin.getPermissionsTool().registerDestinationTeleportPermissions(destination);
```
dumptruckman, line 45: This is great. Creating permissions should not be the concern of DestinationsProvider.

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationsProvider.java
```diff
@@ -65,8 +56,8 @@ private void registerDestinationPerms(@NotNull Destination<?> destination) {
                                                            @Nullable String deststring
     ) {
         return destinationMap.values().stream()
-                .filter(destination -> issuer.hasPermission(PERMISSION_PREFIX + "self." + destination.getIdentifier())
-                        || issuer.hasPermission(PERMISSION_PREFIX + "other." + destination.getIdentifier()))
+                .filter(destination -> this.plugin.getPermissionsTool().hasDestinationTeleportPermission(null, issuer.getIssuer(), destination)
+                        || this.plugin.getPermissionsTool().hasDestinationTeleportPermission(issuer.getIssuer(), this.plugin.getServer().getConsoleSender(), destination))
```
dumptruckman, line 60: Would it make sense to have an overloaded method that takes 1 less param? Also, would it make sense to move the boolean logic into method? (Pretend I don't know what `hasDestinationTeleportPermission` does - which atm is true)

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -105,26 +117,32 @@ private Location getMostAccurateRespawnLocation(World w) {
...
             }
+            Logging.fine("Moving NEW player to(firstspawnoverride): " + worldManager.getFirstSpawnWorld().getSpawnLocation());
+            this.sendPlayerToDefaultWorld(player);
+            return;
+        }
+
+        Logging.finer("Player joined AGAIN!");
+        MVWorld world = worldManager.getMVWorld(player.getWorld());
+        if (world == null) {
+            Logging.fine("Player is in a world that is not managed by Multiverse. Ignoring.");
```
dumptruckman, line 134: This looks like new behavior. Is that right?

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -30,26 +39,29 @@
...
+        actionChecker = plugin.getPlayerActionChecker();
     }
 
     /**
      * @return the playerWorld-map
      */
+    @Deprecated
     public Map<String, String> getPlayerWorld() {
         return playerWorld;
     }
```
dumptruckman, line 67: What was this for and what is replacing it?

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -144,87 +167,54 @@ public void playerChangedWorld(PlayerChangedWorldEvent event) {
...
+
+        if (Objects.equals(teleporterName, "CONSOLE")) {
+            Logging.finer("We know the teleporter is the console! Magical!");
+            teleporter = this.plugin.getServer().getConsoleSender();
+        } else {
+            teleporter = this.plugin.getServer().getPlayerExact(teleporterName);
         }
-        // TODO: Refactor these lines.
-        // Charge the teleporter
-        event.setCancelled(!pt.playerHasMoneyToEnter(fromWorld, toWorld, teleporter, teleportee, true));
```
dumptruckman, line 181: Was this charging the player and also the cause of the double charge bug?

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +243,140 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
+            issuer.sendMessage(targetName + " cannot be teleported to because the destination is null.");
+            return;
+        }
+        if (result.hasResult(NullPlaceResult.NULL_LOCATION)) {
+            issuer.sendMessage(targetName + " cannot be teleported to because the location is null.");
+            return;
+        }
+        if (result.hasResult(NullPlaceResult.NULL_WORLD)) {
+            issuer.sendMessage(targetName + " cannot be teleported because the world is null.");
+            return;
```
dumptruckman, line 358: Why do these first 3 return but the last ones don't?

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +243,140 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
+        if (result.hasResult(NullPlaceResult.NULL_DESTINATION)) {
+            issuer.sendMessage(targetName + " cannot be teleported to because the destination is null.");
+            return;
+        }
+        if (result.hasResult(NullPlaceResult.NULL_LOCATION)) {
+            issuer.sendMessage(targetName + " cannot be teleported to because the location is null.");
+            return;
+        }
+        if (result.hasResult(NullPlaceResult.NULL_WORLD)) {
+            issuer.sendMessage(targetName + " cannot be teleported because the world is null.");
```
dumptruckman, line 357: Will this happen when the world is not managed by MV?

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +243,140 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
+            return;
+        }
+
+        if (result.hasResult(WorldAccessResult.NO_WORLD_ACCESS)) {
+            issuer.sendMessage(targetName + " cannot be teleported to because you does not have access to " + toWorld.getName());
+        }
+        if (result.hasResult(PlayerLimitResult.EXCEED_PLAYERLIMIT)) {
+            issuer.sendMessage(targetName + " cannot cannot enter " + toWorld.getName() + " because it is full.");
```
dumptruckman, line 365:
```suggestion
            issuer.sendMessage(targetName + " cannot enter " + toWorld.getName() + " because it is full.");
```

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -144,87 +167,54 @@ public void playerChangedWorld(PlayerChangedWorldEvent event) {
...
         }
 
-        // By this point anything cancelling the event has returned on the method, meaning the teleport is a success \o/
-        this.stateSuccess(teleportee.getName(), toWorld.getAlias());
-    }
+        if (actionCheckResponse.hasResult(EntryFeeResult.ENOUGH_MONEY)) {
+            double price = toWorld.getPrice();
+            Material currency = toWorld.getCurrency();
+            this.plugin.getEconomist().payEntryFee((Player) teleporter, price, currency);
+        }
```
dumptruckman, line 215: Kind of wondering if we should move this to a separate MONITOR version of this event so we know that the player has actually been teleported.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +243,140 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
-        }
-        event.setCancelled(!pt.playerHasMoneyToEnter(fromWorld, toWorld, event.getPlayer(), event.getPlayer(), true));
-        if (event.isCancelled()) {
-            Logging.fine("Player '" + event.getPlayer().getName()
-                    + "' was DENIED ACCESS to '" + event.getTo().getWorld().getName()
-                    + "' because they don't have the FUNDS required to enter.");
+
+        MVWorld fromWorld = this.worldManager.getMVWorld(event.getFrom().getWorld());
+        MVWorld toWorld = this.worldManager.getMVWorld(event.getTo().getWorld());
+        ActionResponse actionCheckResponse = this.actionChecker.canGoFromToWorld(event.getPlayer(), event.getPlayer(), fromWorld, toWorld);
```
dumptruckman, line 259: I think this action checker method might read better as `canGoFromWorldToWorld`.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +243,140 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
+        Logging.fine(actionCheckResponse.toString());
+        if (!actionCheckResponse.isSuccessful()) {
```
dumptruckman, line 261: Just thought about this here... I think we should generally be looking to avoid negated logic. We can do this by adding a `actionCheckResponse.isFailure`.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +243,140 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
+     * Applies the gamemode and flight for the specified {@link Player}.
+     *
+     * @param player The {@link Player}.
+     * @param world The world the player is in.
+     */
+    public void applyGameModeAndFlight(@NotNull Player player, @NotNull MVWorld world) {
+        if (MVPlayerListener.this.actionChecker.canKeepGameMode(player, world).isSuccessful()) {
+            Logging.fine("Player: " + player.getName() + " is IMMUNE to gamemode changes!");
+            return;
+        }
```
dumptruckman, line 303: Could a reason for keeping game mode also be that the world's game mode isn't different from the player's current game mode? If this is meant to imply something else maybe the action check method could be named to reflect that better.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +243,140 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
+
+        // Check that the player is in the new world and they haven't been teleported elsewhere or the event cancelled.
+        if (player.getWorld() != world.getCBWorld()) {
+            Logging.fine("The gamemode/allowfly was NOT changed for player '%s' because he is now in world '%s' instead of world '%s'",
+                    player.getName(), player.getWorld().getName(), world.getName());
+            return;
+        }
+        Logging.fine("Handling gamemode for player: %s, Changing to %s", player.getName(), world.getGameMode().toString());
+        Logging.finest("From World: %s", player.getWorld());
+        Logging.finest("To World: %s", world);
```
dumptruckman, line 313: Might should be combined into a single message.

src/main/java/com/onarandombox/MultiverseCore/commands/TeleportCommand.java
benwoo1110: Also, canGoToDestination is already checked in the player teleport event.

src/main/java/com/onarandombox/MultiverseCore/destination/DestinationsProvider.java
benwoo1110: lol i actually have an overload alr, smh didnt use it

src/main/java/com/onarandombox/MultiverseCore/api/action/ActionResponse.java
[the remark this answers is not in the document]
dumptruckman: Yes, in which case it wouldn't only find it if it's the exact same object. As long as that is what you want then you can ignore my comment.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: based on the previous code, there was no such check of whether the gamemode is the same, but ig I can add it.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: We could do that yea, but i think not the scope of this PR. Prob do it when i redo the teleport queue stuff, which is really bad. Its like static methods in MultiverseCore class

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: hmm right... its suppose to be `NOT_MV_WORLD` and returns successful

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: yea it was, bcu for some reason the method `playerHasMoneyToEnter` also contain the actual charging of the player, which confused me.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -133,9 +151,14 @@ public void playerJoin(PlayerJoinEvent event) {
...
     @EventHandler(priority = EventPriority.MONITOR)
     public void playerChangedWorld(PlayerChangedWorldEvent event) {
-        // Permissions now determine whether or not to handle a gamemode.
-        this.handleGameModeAndFlight(event.getPlayer(), event.getPlayer().getWorld());
-        playerWorld.put(event.getPlayer().getName(), event.getPlayer().getWorld().getName());
+        MVWorld world = worldManager.getMVWorld(event.getPlayer().getWorld());
+        if (world == null) {
+            Logging.fine("Player is in a world that is not managed by Multiverse. Ignoring.");
+            return;
+        }
```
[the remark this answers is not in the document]
benwoo1110: image

It is actually not a new behvaiour, I just shifted the logic from `handleGameModeAndFlight` method to here instead.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: I dont get the need for tracking what world the player is in, when you can literally do Player#getWorld(). I believe its some legacy thing that should be removed.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: yea, but it makes more sense bcu access permission isnt checked when a player teleports to a non-mv world (even before this PR). So I don't see why MV should check for access permission of a non-mv world on player join.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
dumptruckman: That's fair I guess. It is a public method though, so it might be in use in other plugins. I guess if it is a problem, then this method should have been named better because if it's what it says on the tin then they can just use the method you said.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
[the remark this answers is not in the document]
dumptruckman: Oh, I was looking for the message somewhere else and when I didn't see it I thought this was new.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
dumptruckman: Not quite sure I understand this response... I might be missing the context at this point D:

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
dumptruckman: Want to track this plan somewhere and then we can resolve this convo?

src/main/java/com/onarandombox/MultiverseCore/commands/CheckCommand.java
```diff
@@ -35,6 +35,11 @@ public void onCheckCommand(BukkitCommandIssuer issuer,
     ) {
         issuer.sendMessage("Checking " + player + " to " + destination + "...");
         //TODO More detailed output on permissions required.
-        this.plugin.getDestinationsProvider().checkTeleportPermissions(issuer, player, destination);
+        if (this.plugin.getPlayerActionChecker().canUseDestinationToTeleport(issuer.getIssuer(), player, destination).isUnsuccessful()) {
+            issuer.sendMessage("You don't have permission to use this destination.");
+        }
+        if (this.plugin.getPlayerActionChecker().canGoToDestination(issuer.getIssuer(), player, destination).isUnsuccessful()) {
+            issuer.sendMessage("You don't have permission to teleport to this destination.");
+        }
```
dumptruckman, line 43: Looking at this again I feel like readability is a bit tough. Can we shorten these if statements to just `if (canUseDestinationToTeleport(...)) {`

src/main/java/com/onarandombox/MultiverseCore/commands/TeleportCommand.java
```diff
@@ -35,6 +35,10 @@ public void onTeleportCommand(BukkitCommandIssuer issuer,
     ) {
         // TODO Add warning if teleporting too many players at once.
         for (Player player : players) {
+            if (this.plugin.getPlayerActionChecker().canUseDestinationToTeleport(issuer.getIssuer(), player, destination).isUnsuccessful()) {
```
dumptruckman, line 38: Same thing here. Private helper methods for checks would help with readability.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
```diff
@@ -253,112 +250,136 @@ public void playerPortalCheck(PlayerPortalEvent event) {
...
+    ) {
+        BukkitCommandIssuer issuer = this.plugin.getMVCommandManager().getCommandIssuer(sender);
+        String targetName = issuer.getIssuer() == teleportee ? "You" : teleportee.getName();
+
+        if (result.hasResult(NullPlaceResult.NULL_DESTINATION)) {
+            issuer.sendMessage(targetName + " cannot be teleported to because the destination is null.");
+        }
+        if (result.hasResult(NullPlaceResult.NULL_LOCATION)) {
+            issuer.sendMessage(targetName + " cannot be teleported to because the location is null.");
+        }
```
dumptruckman, line 360: I'm not crazy about all these if statements. I wonder if there's something we could do with polymorphism to clean this up a bit. Would be nice to not have all these extra lines in this class too.

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
```diff
@@ -0,0 +1,329 @@
...
+import org.bukkit.permissions.Permission;
+import org.bukkit.permissions.PermissionDefault;
+import org.bukkit.plugin.PluginManager;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A utility class for registering and checking permissions.
+ */
+public class PermissionsTool {
```
dumptruckman, line 18: Is this changing how permissions get registered and such? Can you summarize how this is different?

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
```diff
@@ -0,0 +1,329 @@
...
+    private Permission bypassPlayerLimitPermission;
+
+    public PermissionsTool(MultiverseCore plugin) {
+        this.plugin = plugin;
+    }
+
+    /**
+     * Sets up the permissions for Multiverse.
+     */
+    public void setUpPermissions() {
```
dumptruckman, line 35: I wonder if this should also be `registerPermissions` since it's registering permissions just like some of the other public methods in this class. Maybe the plugin manager bit belongs elsewhere? Or `setUpWorldPermissionsWildcards` could be public `registerWildcardWorldPermissions()`?

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
```diff
@@ -0,0 +1,329 @@
...
+        this.bypassEntryFeePermission = new Permission("multiverse.exempt.*",
+                "A player who has this does not pay to enter this world, or use any MV portals in it",
+                PermissionDefault.OP);
+        this.bypassPlayerLimitPermission = new Permission("mv.bypass.playerlimit.*",
+                "A player who can enter this world regardless of whether its full",
+                PermissionDefault.OP);
+
+        try {
+            pluginManager.addPermission(this.worldAccessPermission);
+            pluginManager.addPermission(this.bypassGameModePermission);
```
dumptruckman, line 59: I think all these calls to method calls to methods on `pluginManager` could maybe be move into some private helper methods in here to DRY this up just a bit.

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
```diff
@@ -0,0 +1,329 @@
...
+
+    /**
+     * Registers the permissions for a world.
+     *
+     * @param world The world to register permissions for.
+     */
+    public void registerMVWorldPermissions(MVWorld world) {
+        Permission accessPermission = new Permission("multiverse.access." + world.getName(),
+                "Allows access to " + world.getName(),
+                PermissionDefault.OP);
```
dumptruckman, line 77: Same DRYing could be done with creation of permissions, methods to reduce the repeated stuff in these lines.

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
```diff
@@ -0,0 +1,329 @@
...
+    /**
+     * Checks if a player has the permission to bypass the player limit.
+     *
+     * @param sender    The player to check.
+     * @param toWorld   The world to check.
+     * @return True if the player has the permission, false otherwise.
+     */
+    public boolean hasBypassPlayerLimit(@NotNull CommandSender sender, @NotNull MVWorld toWorld) {
+        return hasPermission(sender, "mv.bypass.playerlimit." + toWorld.getName());
+    }
```
dumptruckman, line 205: I think we should probably split permission checking into a separate class from the class doing permission registration. I wonder, could all the registration stuff be package-private so it's not part of our exported API? This this checking stuff could be in PermissionChecker or something?

src/main/java/com/onarandombox/MultiverseCore/utils/player/PlayerActionChecker.java
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * Checks if a player have permissions to teleport to a destination by another teleporter.
+     *
+     * @param teleporter    The player who is teleporting the teleportee.
+     * @param teleportee    The player to teleport.
+     * @param destination   The destination to teleport to.
+     * @return The result of the check.
+     */
+    public UseDestinationResult canUseDestinationToTeleport(@NotNull CommandSender teleporter,
```
dumptruckman, line 53: Why are some of the return types here `Result` and some `Response`?

src/main/java/com/onarandombox/MultiverseCore/utils/player/checkresult/BlacklistResult.java
```diff
@@ -0,0 +1,32 @@
+package com.onarandombox.MultiverseCore.utils.player.checkresult;
+
+import com.onarandombox.MultiverseCore.api.action.SimpleActionResult;
+
+public class BlacklistResult extends SimpleActionResult {
```
dumptruckman, line 5: Hmm... I'm getting kind of hung up on these class names. BlacklistResult makes me think there is some kind of blacklist action that we're getting the result for. But it's not a blacklist action, it's a permission check action or similar we're trying to get the result for.

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorld.java
```diff
@@ -384,37 +377,6 @@ public String toString() {
         };
     }
 
-    /**
-     * Initializes permissions.
-     */
-    private void initPerms() {
```
dumptruckman, line 390: Ooh yeah. This is not a world concern. Nice.

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorldManager.java
```diff
@@ -722,29 +720,10 @@ public void loadWorlds(boolean forceLoad) {
...
-                }
-                this.plugin.getServer().getPluginManager().removePermission(w.getAccessPermission().getName());
-                this.plugin.getServer().getPluginManager().removePermission(w.getExemptPermission().getName());
-                // Special namespace for gamemodes
-                this.plugin.getServer().getPluginManager().removePermission("mv.bypass.gamemode." + w.getName());
-            }
-            // Recalc the all permission
-            this.plugin.getServer().getPluginManager().recalculatePermissionDefaults(allAccess);
-            this.plugin.getServer().getPluginManager().recalculatePermissionDefaults(allExempt);
+            this.plugin.getPermissionsTool().removeAllMVWorldPermissions();
```
dumptruckman, line 726: Love to see it.

src/main/java/com/onarandombox/MultiverseCore/utils/player/checkresult/BlacklistResult.java
benwoo1110: actually yea that is possible

src/main/java/com/onarandombox/MultiverseCore/utils/player/PlayerActionChecker.java
benwoo1110: Response is multiple `ActionResult` object chained, see `ActionResponse` class. Main purpose is so we can get more details of the check and not just a boolean.

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
benwoo1110: yea can

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
benwoo1110: not really, its just refactoring the registering of permission to bukkit api from being all scattered between different classes into 1 class.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: ig setting up github project maybe?

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: `NULL_WORLD` is when the bukkit world object is null, if world is just not managed by MV, it should be `NOT_MV_WORLD`, which does this:
image

which is the same behaviour as before, just refactored

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: yea, will address this issue with a seperate PR. Either ways the API will be breaking, so it will either be refactor out of playerListener class or removed if I still cant see the purpose of it.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
dumptruckman: I see. Would a name like `NOT_BUKKIT_WORLD` or something make sense?

benwoo1110: Uh if it's not a bukkit world, means it's null, there isn't such a thing as a non-bukkit world. And then MVWorld is kinda just an extension of bukkit world

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
dumptruckman: https://github.com/orgs/Multiverse/projects/4/views/1?pane=issue&itemId=21658559

src/main/java/com/onarandombox/MultiverseCore/utils/player/PlayerActionChecker.java
dumptruckman: That relationship is hard to notice from a distance. Could we maybe use a name like `ActionResultChain` or something that makes that notion clearer?

dumptruckman: Ultimately though, I think the best case would be to just think about `ActionResult` and if it chains, that should hidden in the implementation unless there's a really good reason that other things need to know that the result is a chain or results rather than just if it's successful or not.

dumptruckman:
Looks like maybe you rebased this branch and force pushed. Just FYI that makes it harder to tell what has changed since I last looked at this. I can't easily look at the commits since my last comment and now just have to all files that have changed since I last looked at this.

src/main/java/com/onarandombox/MultiverseCore/commands/CheckCommand.java
dumptruckman: Musing about this again, sorry... I feel that just in general `canUseDestinationToTeleport` feels like it should return a boolean. When I see `canUseDestinationToTeleport.isUnsuccessful` or `canUseDestinationToTeleport.isSuccessful` it strikes me as strange.

benwoo1110: I remove isUnsuccessful and isSuccessful, and changed them to just asBoolean.

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: nice, will add to the project page

src/main/java/com/onarandombox/MultiverseCore/utils/player/checkresult/BlacklistResult.java
benwoo1110: But thinking again, I think having a specific return type like `BlacklistResult` is better than `OperationResult`, so users know exactly only these possible results will be returned, not any other results.

src/main/java/com/onarandombox/MultiverseCore/utils/player/PlayerActionChecker.java
benwoo1110: I think it makes it clearer that this function will return more than 1 result, vs another function that will only return 1 result.

But i do agree with the class naming, made that changed alr.

src/main/java/com/onarandombox/MultiverseCore/utils/PermissionsTool.java
benwoo1110: hm but i realised `package-private` wont work tho, bcu its in the `utils` package, and it will need to be access by other parent packages

src/main/java/com/onarandombox/MultiverseCore/utils/player/checkresult/BlacklistResult.java
dumptruckman: Hmm yeah, that's fair.

nicegamer7 (approved):

src/main/java/com/onarandombox/MultiverseCore/listeners/MVPlayerListener.java
benwoo1110: yea prob some form of mapping, but I think its fine for now. prob another thing into the todo bucket

src/main/java/com/onarandombox/MultiverseCore/anchor/AnchorManager.java
```diff
@@ -142,7 +142,7 @@ public Set<String> getAnchors(Player p) {
             }
             MVWorld mvWorld = this.plugin.getMVWorldManager().getMVWorld(ancLoc.getWorld());
             if (mvWorld != null) {
-                if (this.plugin.getPlayerActionChecker().hasAccessToWorld(p, mvWorld).isUnsuccessful()) {
+                if (!this.plugin.getPlayerActionChecker().hasAccessToWorld(p, mvWorld).asBoolean()) {
```
dumptruckman, line 145: This isn't really what I meant. Something like `if (!hasAnchorAccessInWorld(player, mvWorld)) {`
```java
private boolean hasAnchorAccessInWorld(Player player, MVWorld mvWorld) {
    return plugin.getPlayerActionChecker().hasAccessToWorld(player, mvWorld).isSuccessful();
}
```
Btw... is this about anchor access in a world cause that's what the logged message says. However, the code doesn't seem to indicate that.

src/main/java/com/onarandombox/MultiverseCore/api/operation/SimpleOperationResult.java
```diff
@@ -34,22 +34,22 @@ public static ActionResult forFailure(String name) {
...
      */
-    public static ActionResult of(String name, boolean isSuccessful) {
-        return new SimpleActionResult(name, isSuccessful);
+    public static OperationResult of(String name, boolean isSuccessful) {
+        return new SimpleOperationResult(name, isSuccessful);
     }
 
     private final String name;
-    private final boolean isSuccessful;
+    private final boolean booleanState;
```
dumptruckman, line 42: I think `isSuccessful` is way more meaningful.

src/main/java/com/onarandombox/MultiverseCore/utils/permission/PrefixPermission.java
```diff
@@ -0,0 +1,102 @@
+package com.onarandombox.MultiverseCore.utils.permission;
+
+import org.bukkit.Bukkit;
+import org.bukkit.permissions.Permission;
+import org.bukkit.permissions.PermissionDefault;
+import org.bukkit.plugin.PluginManager;
+
+public class PrefixPermission {
```
dumptruckman, line 9: What is PrefixPermission and why is it also registering permissions? Shouldn't the PermissionRegistrar be doing that?

src/main/java/com/onarandombox/MultiverseCore/api/operation/SimpleOperationResult.java
benwoo1110: ok ig i will change back then

src/main/java/com/onarandombox/MultiverseCore/anchor/AnchorManager.java
benwoo1110: ahh

src/main/java/com/onarandombox/MultiverseCore/utils/permission/PrefixPermission.java
benwoo1110: PrefixPermission is so that you can dynamically get like `mv.bypass.gamemode.` then you do `PrefixPermission.getPermissionString(<worldname>)` to get `mv.bypass.gamemode.<worldname>`, which is better than using string concatenation

src/main/java/com/onarandombox/MultiverseCore/utils/permission/PermissionsRegistrar.java
```diff
@@ -21,15 +21,15 @@ public class PermissionsRegistrar {
...
         worldPermissions = new ArrayList<>() {{
-            worldAccessPermission = new PrefixPermission("multiverse.access.", "Allows access to a world.");
-            worldGamemodeBypassPermission = new PrefixPermission("mv.bypass.gamemode.", "Allows bypassing of gamemode restrictions.", PermissionDefault.FALSE);
-            worldExemptPermission = new PrefixPermission("multiverse.exempt.", "A player who has this does not pay to enter this world.");
-            worldPlayerLimitBypassPermission = new PrefixPermission("mv.bypass.playerlimit.", "Allows bypassing of player limit restrictions.");
+            add(worldAccessPermission = new PrefixPermission("multiverse.access.", "Allows access to a world."));
+            add(worldGamemodeBypassPermission = new PrefixPermission("mv.bypass.gamemode.", "Allows bypassing of gamemode restrictions.", PermissionDefault.FALSE));
+            add(worldExemptPermission = new PrefixPermission("multiverse.exempt.", "A player who has this does not pay to enter this world."));
+            add(worldPlayerLimitBypassPermission = new PrefixPermission("mv.bypass.playerlimit.", "Allows bypassing of player limit restrictions."));
         }};
```
dumptruckman, line 28: I think we should avoid double brace initialization as it creates a subclass of ArrayList with an instance initializer which is generally overkill for just about any use case. Also, assignment inside a method call should be avoided. Each assignment should always be on its own line.

PR #2839, opened 2023-02-12
feat: Implement remove and import command

benwoo1110:
Implement import and remove command


src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
```diff
@@ -0,0 +1,88 @@
...
+    @CommandPermission("multiverse.core.import")
+    @CommandCompletion("@mvworlds:scope=potential  @flags:groupName=mvimport")
+    @Syntax("<name> <env> --generator [generator[:id]] --adjust-spawn")
+    @Description("Imports a existing world folder.")
+    public void onImportCommand(BukkitCommandIssuer issuer,
+
+                                @Conditions("validWorldName:scope=new")
+                                @Syntax("<name>")
+                                @Description("Folder name of the world.")
+                                String worldName,
```
zax71, line 60: `folderName` makes more sense to me

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
```diff
@@ -0,0 +1,88 @@
...
+                                World.Environment environment,
+
+                                @Optional
+                                @Syntax("--generator [generator[:id]] --adjust-spawn")
+                                @Description("Other world settings. See: http://gg.gg/nn8c2")
+                                String[] flags) {
+
+        ParsedCommandFlags parsedFlags = parseFlags(flags);
+
+        issuer.sendMessage(String.format("Starting import of world '%s'...", worldName));
```
zax71, line 73: No need to say `world <worldName>`, would be better as
```java
issuer.sendMessage(ChatColor.GREEN + "Starting import of " + ChatColor.GOLD + worldName);
```

zax71 (approved):
Again, just UX stuff

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
benwoo1110: I mean the folder name is the world name tbh, but ig can change to differentiate from create command

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
benwoo1110: hmm yea, tbh we need to standardise our output formatting colour and stuff

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
dumptruckman: I dunno, I'm not sure that is significant enough to warrant changing this from prior expectations of general terms we use.

benwoo1110: yea, users wont see our variables anyways

dumptruckman: Maybe the description would read better like "Name of the world folder".

src/main/java/com/onarandombox/MultiverseCore/commands/ImportCommand.java
dumptruckman: I have given thought to that problem before but had trouble coming up with a good solution. It's hard to come up with a concise list of "categories" to assign colors to.

benwoo1110: yea, i see other plugins most colours are hard coded also

benwoo1110: Currently not too concerned about how fancy the output is, will be moved to locales anyways

PR #2842, opened 2023-02-17
feat: Improve locale support

benwoo1110:
- Adds loading custom.properties files from the plugin folder
- Enable per-player locale support


src/main/java/com/onarandombox/MultiverseCore/locale/FileResClassLoader.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import org.bukkit.plugin.Plugin;
+
+/**
+ * A class loader that loads resources from the plugin's locales folder.
+ */
+public class FileResClassLoader extends ClassLoader {
```
dumptruckman, line 15: Maybe name this something locale related since it is specifically for that.

dumptruckman (approved):

PR #2859, opened 2023-02-22
Add maven central to repositories

benwoo1110:
Some dependencies can be directly obtained from maven central, so good to add it as a repo.


dumptruckman:
This needs to be a patch release because the _latest_ release was a full release. Otherwise we end up making 4.3.8-pre when 4.3.8 is already out. This is one thing that could be better about the release process...

nicegamer7 (approved):

PR #2880, opened 2023-03-02
feat: Add automatic Modrinth upload

zax71:
* Adds the `modrinth` Gradle task
* Modifies `release.yml` to automatically release to Modrinth


build.gradle
```diff
@@ -171,3 +172,20 @@ shadowJar {
...
+    versionNumber = version
+
+    if(System.getenv("PRERELEASE")) {
+        versionType = "beta"
+    } else {
+        versionType = "release"
+    }
+
+    uploadFile = jar
+    gameVersions = ["1.19.2", "1.19.1", "1.19", "1.18.2", "1.18.1", "1.18", "1.17.1", "1.17", "1.16.5", "1.16.4", "1.16.3", "1.16.2", "1.16.1", "1.16"]
```
benwoo1110, line 188: Since this targets the main, we should add versions 1.13-1.15

build.gradle
```diff
@@ -171,3 +172,20 @@ shadowJar {
...
+    loaders = ["bukkit", "spigot", "paper", "purpur"]
```
benwoo1110, line 189: not very sure if we should explicitly put purpur hm

zax71: Welp, It does support it, so I don't really see why not

build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
     id 'maven-publish'
     id 'checkstyle'
     id 'com.github.johnrengelman.shadow' version '7.1.2'
+    id "com.modrinth.minotaur" version "2.+"
```
benwoo1110, line 8: i think exact version is better

zax71: From Modrinth:
> It's safest to have this on 2.+ to get the latest features and bug fixes without having to worry about breaking changes.

benwoo1110: That is if you trust people code actually don't break stuff hehe

zax71: I think we can trust Modrinth xD

benwoo1110 (approved):

dumptruckman (approved):

PR #2888, opened 2023-03-12
Add PlaceholderAPI Support

zax71:


src/main/java/com/onarandombox/MultiverseCore/MultiverseCorePlaceholders.java
```diff
@@ -0,0 +1,117 @@
...
+    private final MultiverseCore plugin;
+
+
+    public MultiverseCorePlaceholders(MultiverseCore plugin) {
+        this.plugin = plugin;
+    }
+
+    @Override
+    public @NotNull String getIdentifier() {
+        return "multiverse";
```
benwoo1110, line 23: `multiverse-core` may be better, bcu we may have papi extension for our other submodules

darbyjack:
Note to self: Unverify the eCloud MV expansion when this gets merged for release so that we don't have players getting confused.

benwoo1110:
I was also thinking of adding support for like specific worldname, instead of just based on the one player is in. Example %multiverse_name_<worldname>% or similar.

And will need a config option for it, but only after my commented config changes, so will need to add in our mv5 project todo.

benwoo1110:
Oh I mean config option to register mv papi extension or not. Was talking about completely separate 2 pointers.

zax71:
Format is: `%multiverse-core_<world>_<placeholder>` or `%multiverse-core_<placeholder>` and the i18n support is half-baked, need a hand with the rest if you can Ben (see Discord)

src/main/java/com/onarandombox/MultiverseCore/placeholders/MultiverseCorePlaceholders.java
```diff
@@ -44,116 +46,107 @@ public boolean persist() {
...
+     * Placeholder implementation, format: %multiverse-core_&lt;placeholder&gt;_[world]% world is optional.
+     *
+     * @param offlinePlayer Player to get the placeholder for
+     * @param params        Placeholder to get
+     * @return Placeholder value
      */
     @Override
     public @Nullable String onRequest(OfflinePlayer offlinePlayer, @NotNull String params) {
-        final MVWorldManager worldManager = plugin.getMVWorldManager();
+        // Split string in to an Array wt the underscores
```
zax71, line 58: "Array wt the underscores"

benwoo1110 (dismissed):

PR #2895, opened 2023-03-22
Revamp config

benwoo1110:
Checklist:
- [x] mv config command
- [x] Node value validation

https://github.com/Multiverse/Multiverse-Core/suites/11833308794/artifacts/618148459


src/main/java/com/onarandombox/MultiverseCore/configuration/DefaultMVConfig.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Override
+    public void setPortalSearchRadius(int searchRadius) {
+        settings.set(MVConfigNodes.CUSTOM_PORTAL_SEARCH_RADIUS, searchRadius);
+    }
+
+    @Override
+    public int getPortalSearchRadius() {
+        return settings.get(MVConfigNodes.CUSTOM_PORTAL_SEARCH_RADIUS);
+    }
```
dumptruckman, line 146: Crazy idea time.

I was reflecting on using more specific types to constrain allowable values. For example, having a type like PortalSearchRadius that can only represent valid values for it's use case. I guess that's maybe just 0 or higher for this case?

If we did use something like that, we could have a config system that could do something if the value isn't acceptable. For example, if this was set to -5, log a warning about this value and say that the default will be used instead. Then when we ask for this config value in the code, it's always going to be an valid value, no error handling needed. And if there is an error, it's likely that the valid value changed, in which case we can potentially only have to update the PortalSearchRadius type.

src/main/java/com/onarandombox/MultiverseCore/configuration/MVConfigNodes.java
```diff
@@ -0,0 +1,170 @@
...
+        nodes.add(node);
+        return node;
+    }
+
+    private static final MVCommentedNode HEADER = node(MVCommentedNode.builder("world") // TODO hacky way to get the header to the top of the file
+            .comment("####################################################################################################")
+            .comment("#                                                                                                  #")
+            .comment("#                    █▀▄▀█ █░█ █░░ ▀█▀ █ █░█ █▀▀ █▀█ █▀ █▀▀   █▀▀ █▀█ █▀█ █▀▀                      #")
+            .comment("#                    █░▀░█ █▄█ █▄▄ ░█░ █ ▀▄▀ ██▄ █▀▄ ▄█ ██▄   █▄▄ █▄█ █▀▄ ██▄                      #")
+            .comment("#                                                                                                  #")
```
dumptruckman, line 27: 😄

src/main/java/com/onarandombox/MultiverseCore/configuration/MVConfigNodes.java
```diff
@@ -0,0 +1,170 @@
...
+            .comment("#    Each option in this file is documented and explained here:                                    #")
+            .comment("#     ==>  https://github.com/Multiverse/Multiverse-Core/wiki/config.yml                           #")
+            .comment("#                                                                                                  #")
+            .comment("#                                                                                                  #")
+            .comment("#    New options are added to this file automatically. If you manually made changes                #")
+            .comment("#    to this file while your server is running, please run `/mv reload` command.                   #")
+            .comment("#                                                                                                  #")
+            .comment("####################################################################################################")
+            .comment("")
+            .comment("")
```
dumptruckman, line 43: Killer header.

src/main/java/com/onarandombox/MultiverseCore/configuration/MVConfigNodes.java
```diff
@@ -0,0 +1,170 @@
...
+//            .comment("")
+//            .build());
+
+    public static final MVValueNode<Boolean> ENFORCE_ACCESS = node(MVValueNode.builder("world.enforce-access", Boolean.class)
+            .comment("This setting will prevent players from entering worlds they don't have access to.")
+            .comment("If this is set to false, players will be able to enter any world they want.")
+            .comment("If this is set to true, players will only be able to enter worlds they have")
+            .comment("the `mv.access.<worldname>` permission.")
+            .defaultValue(false)
+            .build());
```
dumptruckman, line 57: Consider using a lambda function parameter to reduce the need to know about `MVValueNode.builder` and `MVCommentedNode.builder`. For example `valueNode("world.enforce-access", Boolean.class, node -> node.comment...` This removes the need to call `build` at the end as well.

src/main/java/com/onarandombox/MultiverseCore/configuration/MVConfigNodes.java
```diff
@@ -0,0 +1,170 @@
...
+            .comment("")
+            .comment("If true, the startup console messages will no longer show.")
+            .defaultValue(false)
+            .build());
+
+    public static final MVValueNode<Boolean> I_DONT_WANT_TO_DONATE = node(MVValueNode.builder("misc.i-dont-want-to-donate", Boolean.class)
+            .comment("")
+            .comment("If you don't want to donate, you can set this to true and Multiverse will stop nagging you.")
+            .defaultValue(false)
+            .build());
```
dumptruckman, line 161: What if we moved the donation message to the top of the config file instead of logging in onEnable?

src/main/java/com/onarandombox/MultiverseCore/utils/settings/MVSettings.java
```diff
@@ -0,0 +1,232 @@
...
+import com.onarandombox.MultiverseCore.utils.settings.migration.ConfigMigrator;
+import io.github.townyadvanced.commentedconfiguration.CommentedConfiguration;
+import io.github.townyadvanced.commentedconfiguration.setting.CommentedNode;
+import io.github.townyadvanced.commentedconfiguration.setting.TypedValueNode;
+import io.github.townyadvanced.commentedconfiguration.setting.ValueNode;
+import org.bukkit.plugin.Plugin;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MVSettings {
```
dumptruckman, line 18: Not crazy about having a thing called Settings and a thing called Config. How are they different and how do we easily understand that? If `MVConfig` is the primary place to get settings, what if we call this one something like `ConfigHandle`?

src/main/java/com/onarandombox/MultiverseCore/utils/settings/migration/MoveMigratorAction.java
```diff
@@ -0,0 +1,31 @@
...
+        this.fromPath = fromPath;
+        this.toPath = toPath;
+    }
+
+    @Override
+    public void migrate(MVSettings settings) {
+        Logging.info(String.valueOf(settings.getConfig().get(fromPath)));
+        Optional.ofNullable(settings.getConfig().get(fromPath))
+                .ifPresent(value -> {
+                    Logging.info("Moving " + fromPath + " to " + toPath);
```
dumptruckman, line 26: Intentionally logging twice?

src/main/java/com/onarandombox/MultiverseCore/utils/settings/node/NodesCollection.java
dumptruckman, line 1: What's up with this wrapper class?

src/main/java/com/onarandombox/MultiverseCore/utils/settings/migration/MoveMigratorAction.java
benwoo1110: Ahh right, added it for debugging with I was struggling to get migration to work. I will reduce the logging in the next commit.

src/main/java/com/onarandombox/MultiverseCore/configuration/MVConfigNodes.java
benwoo1110: Uh that's your call, the donation are to you haha. I'm ok with changing it.

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -378,31 +369,10 @@ public MVWorldManager getMVWorldManager() {
...
-            coreDefaults = YamlConfiguration.loadConfiguration(new BufferedReader(new InputStreamReader(resourceURL, "UTF-8")));
-            this.multiverseConfig.setDefaults(coreDefaults);
-        } catch (UnsupportedEncodingException e) {
-            Logging.severe("Couldn't load default config with UTF-8 encoding. Details follow:");
-            e.printStackTrace();
-            Logging.severe("Default configs NOT loaded.");
-        }
+        config = new DefaultMVConfig(this);
+        config.load();
+        config.save();
```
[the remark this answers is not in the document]
benwoo1110: Will do!

PR #2909, opened 2023-04-10
Revamp perms 2

benwoo1110:


src/main/java/com/onarandombox/MultiverseCore/listeners/NewMVPlayerListener.java
dumptruckman, line 1: Is putting this in a new class really necessary? Would be nice to see a diff to really see what is changing.

src/main/java/com/onarandombox/MultiverseCore/utils/permissions/ChildPermission.java
dumptruckman, line 1: What is the point of this?

src/main/java/com/onarandombox/MultiverseCore/utils/result/Result.java
```diff
@@ -0,0 +1,84 @@
...
+        return new Failure<>(failureReason);
+    }
+
+    boolean isSuccess();
+
+    boolean isFailure();
+
+    S successReason();
+
+    F failureReason();
```
dumptruckman, line 21: Funny thing is, since this is a sealed interface, we could use default implementations that check `instanceof` exhaustively. Not really sure if there's anything better about either way though.

dumptruckman, line 21: Sort of feel that since this isn't a builder/dsl, it would be better to use `getSuccessReason` and `getFailureReason`.

src/main/java/com/onarandombox/MultiverseCore/listeners/NewMVPlayerListener.java
dumptruckman: Alternately, refactoring to shorter methods could help.

src/main/java/com/onarandombox/MultiverseCore/listeners/NewMVPlayerListener.java
```diff
@@ -0,0 +1,326 @@
...
+                .onFailureReason(PlayerLimitResult.Failure.class, reason -> {
+                    teleporter.sendMessage("World '%s' is full".formatted(toWorld.getName()));
+                })
+                .onFailureReason(EntryFeeResult.Failure.class, reason -> {
+                    teleporter.sendMessage("You do not have enough money to pay world '%s' entry fee.".formatted(toWorld.getName()));
+                    //TODO Formatted entry fee amount
+                })
+                .onFailureReason(BlacklistResult.Failure.class, reason -> {
+                    teleporter.sendMessage("World '%s' is blacklisted by world '%s'".formatted(fromWorld.getName(), toWorld.getName()));
+                });
```
dumptruckman, line 277: This is what has been throwing me off. Like, what if a new failure reason is added and used elsewhere and forgotten here? Is there some way we can avoid that kind of situation?

dumptruckman (approved):

PR #2975, opened 2023-08-25
Implement `/mv dumps` command

zax71:
## Current changes/additions

 - Renamed `/mv version` to `/mv dumps`
 - Added `--logs` option that uploads logs to https://mclo.gs
 - Added `bukkit.yml` and `server.properties` to dump
 - Added installed plugin list to dump
 - Implemented `--pastebincom` `--pastesdev` `--pastegg` `--logs` and `--exclude-plugin-list`
 - Replace Hastebin with [pastes.dev](https://pastes.dev)


dumptruckman:
I don't think I'm a fan of completely replacing the version command. I would rather keep that command and maybe change it's functionality - perhaps it logs out some version information or something to the console/sender. Then it could also mention that `/mv dumps` can be used to share information about the plugin/their server.

zax71:
I might add that in another PR, would like to finish this one first though :)

src/main/java/com/onarandombox/MultiverseCore/configuration/node/NodeGroup.java
```diff
@@ -1,15 +1,11 @@
...
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
-
 import io.github.townyadvanced.commentedconfiguration.setting.CommentedNode;
 import io.vavr.control.Option;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.*;
```
benwoo1110, line 7: Don't use wildcard imports, intellij has a config for it. I usually just set like to 9999 imports before using wildcards lol

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorldManager.java
```diff
@@ -104,23 +83,27 @@ public SimpleMVWorldManager(
...
-        File[] files = server.getWorldContainer().listFiles((file, s) -> s.equalsIgnoreCase("bukkit.yml"));
-        if (files != null && files.length == 1) {
-            FileConfiguration bukkitConfig = YamlConfiguration.loadConfiguration(files[0]);
-            if (bukkitConfig.isConfigurationSection("worlds")) {
-                Set<String> keys = bukkitConfig.getConfigurationSection("worlds").getKeys(false);
-                for (String key : keys) {
-                    defaultGens.put(key, bukkitConfig.getString("worlds." + key + ".generator", ""));
-                }
+
+        if (plugin.getBukkitConfig() == null) {
```
benwoo1110, line 94: Is that change relevant to the /mv dumps command? Also I recall you said getBukkitConfig() is only supported on paper or smt?

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorldManager.java
```diff
@@ -1001,24 +985,24 @@ public FileConfiguration getConfigWorlds() {
...
+        }
+        return false;
+    }
 
     /**
      * {@inheritDoc}
      */
     @Override
-	public Collection<String> getPotentialWorlds() {
+    public Collection<String> getPotentialWorlds() {
```
benwoo1110, line 1005: I think all these formatting issue should be a separate pr, I think your IDE prob auto fixed the formatting Bcu you edited the file lol

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,242 @@
...
+                        @NotNull MultiverseCore plugin,
+                        @NotNull MVCoreConfig config,
+                        @NotNull MVWorldManager worldManager) {
+        super(commandManager);
+        this.config = config;
+        this.plugin = plugin;
+        this.worldManager = worldManager;
+
+        registerFlagGroup(CommandFlagGroup.builder("mvdumps")
+                .add(CommandFlag.builder("--pastebincom")
```
benwoo1110, line 53: I think --pastebin good enough, don't need com

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,242 @@
...
+                .add(CommandFlag.builder("--hastebin")
+                        .addAlias("-h")
+                        .build())
+                .add(CommandFlag.builder("--pastegg")
+                        .addAlias("-p")
+                        .build())
+                .add(CommandFlag.builder("--logs")
+                        .addAlias("-l")
+                        .build())
+                .add(CommandFlag.builder("--include-plugin-list")
```
benwoo1110, line 68: I actually had the thought, plugin list should be included by default. And have a flag to exclude it.

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,242 @@
...
+
+        BukkitRunnable logPoster = new BukkitRunnable() {
+            @Override
+            public void run() {
+                HashMap<String, String> pasteURLs = new HashMap<>();
+                boolean hasArgs = false;
+
+                if (parsedFlags.hasFlag("--pastebincom")) {
+                    hasArgs = true;
+                    sendMessage(issuer, "Uploading debug info to https://pastebin.com");
```
benwoo1110, line 110: Prob good idea to make all the sendMessage string into i18n instead

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorldManager.java
zax71: Yep, that was the case lol

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
zax71: I did think that originally but I don’t people using —pagebin to put it on a pastebon, not realising it actually means pastebin.com

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
zax71: Nice idea. I’ll implement that one

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
zax71: Adding to the todo

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorldManager.java
zax71: I needed the Bukkit.yml file and server.properties for the dump so I felt it should be abstracted there too. It also makes it much easier to edit when the issue is fixed

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
benwoo1110: huh then what it means?

zax71: Upload to https://pastebin.com

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -330,6 +330,34 @@ public String getAuthors() {
         return authors.toString();
     }
 
+    @Nullable
+    public File getBukkitConfig() {
```
dumptruckman, line 334: I'm not crazy about adding these methods to MultiverseCore after I worked to move most of our API stuff out of it. Is this functionality that is going to be useful elsewhere? Instead, we could inject the Server into the command class and just make these private methods in there.

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+                    String link = pasteURLs.get(service);
+                    issuer.sendInfo(MVCorei18n.DUMPS_URL_LIST, "{service}", service, "{link}", link);
+                }
+
+            }
+        };
+
+        // Run the uploader async as it could take some time to upload the debug info
+        logPoster.runTaskAsynchronously(plugin);
+    }
```
dumptruckman, line 201: This method is nearly 90 lines long (way too long). Could we extract some of this, like especially the runnable stuff, into a separate method/class?

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorldManager.java
```diff
@@ -103,23 +103,27 @@ public SimpleMVWorldManager(
...
+            Logging.warning("Any Default worldgenerators will not be loaded!");
+            return;
+        }
+
+        FileConfiguration bukkitConfig = YamlConfiguration.loadConfiguration(bukkitConfigFile);
+
+        if (bukkitConfig.isConfigurationSection("worlds")) {
+            Set<String> keys = bukkitConfig.getConfigurationSection("worlds").getKeys(false);
+            for (String key : keys) {
+                defaultGens.put(key, bukkitConfig.getString("worlds." + key + ".generator", ""));
```
dumptruckman, line 125: We might want to wait until this is dealt with in the new world manager...

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+
+        // Deal with default case of upload
+        Services services = parsedFlags.flagValue("--upload", Services.class);
+        if (services == null) {
+            services = Services.PASTEGG;
+            Logging.finer("Upload was null so set to pastegg");
+        }
+
+
+        // Need to be final for some reason...
```
benwoo1110, line 152: This is bcu you reference the variable in anonymous function

zax71: Ah, ty for clearing that up!

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+                        pasteURLs.put("paste.gg", postToService(PasteServiceType.PASTEGG, true, null, files));
+                    }
+                    case PASTESDEV -> {
+                        issuer.sendInfo(MVCorei18n.DUMPS_UPLOADING, "{link}", "https://pastes.dev");
+                        pasteURLs.put("pastes.dev", postToService(PasteServiceType.PASTESDEV, true, null, files));
+                    }
+                }
+
+                // Finally, loop through and print all URLs
+                for (String service : pasteURLs.keySet()) {
```
benwoo1110, line 191: I dont really see a need for loop, bcu its just printing 1 or 2 things

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+        final Services finalServices = services;
+        BukkitRunnable logPoster = new BukkitRunnable() {
+            @Override
+            public void run() {
+                HashMap<String, String> pasteURLs = new HashMap<>();
+                Logging.finer("Logs type is: " + finalLogsType);
+                Logging.finer("Services is: " + finalServices);
+
+                // Deal with logs flag
+                if (!parsedFlags.hasFlag("--paranoid")) {
```
benwoo1110, line 163: I think this check should be done above with the `// Deal with default case of logs`

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+                        .build())
+                .build());
+    }
+
+    private enum Services {
+        PASTEGG,
+        PASTESDEV
+    }
+
+    private enum LogsType {
```
benwoo1110, line 105: `LogsTypeOption` maybe is clearer?

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+                            return types;
+                        })
+                        .build())
+                .add(CommandFlag.builder("--paranoid")// Does not upload logs or plugin list (except if --logs mclogs is there)
+                        .addAlias("-p")
+                        .build())
+                .build());
+    }
+
+    private enum Services {
```
benwoo1110, line 100: `SevicesTypeOption` maybe is clearer?

benwoo1110 (changes requested):

src/main/java/com/onarandombox/MultiverseCore/world/SimpleMVWorldManager.java
benwoo1110: I think its fine,  new world manager wont be touching generators stuff anytime soon

src/main/java/com/onarandombox/MultiverseCore/utils/MVCorei18n.java
```diff
@@ -37,6 +37,11 @@ public enum MVCorei18n implements MessageKeyProvider {
     DELETE_SUCCESS,
     DELETE_PROMPT,
 
+    // Dumps command
+    DUMPS_DESCRIPTION,
+    DUMPS_UPLOADING,
+    DUMPS_UPLOADING_LOGS,
+    DUMPS_URL_LIST,
```
benwoo1110, line 44: new line after this is preferred

src/main/java/com/onarandombox/MultiverseCore/configuration/node/NodeGroup.java
```diff
@@ -10,6 +10,7 @@
 import java.util.Iterator;
 import java.util.Map;
 
+
```
benwoo1110, line 13: why a random new line?

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+
+        // Deal with default case of logs
+        LogsType logsType = parsedFlags.flagValue("--logs", LogsType.class);
+        if (logsType == null) {
+            logsType = LogsType.MCLOGS;
+            Logging.finer("Logs was null so set to mclogs");
+        }
+
+
+
```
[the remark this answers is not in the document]
zax71: Sorry... I always forget to fix it lol

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
zax71: Good point. I might just make it print out directly

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,320 @@
...
+
+        // Deal with default case of upload
+        Services services = parsedFlags.flagValue("--upload", Services.class);
+        if (services == null) {
```
benwoo1110, line 144: set a defaultValue for command flag, so you dont need null checking

benwoo1110 (commented):
I think overall the logic checking can be more readable instead of scatter at various parts of the code, for example don't reuse hasFlag for the same flag multiple times:

```java
boolean paranoid = parsedFlags.hasFlag("--paranoid");
Services services = parsedFlags.flagValue("--upload", Services.class);
LogsType logsType = parsedFlags.flagValue("--logs", LogsType.class);
if (!paranoid && logsType == null) {
     logsType = LogsType.MCLOGS;
}
```

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
zax71: Regarding moving `logPoster` to its own class, where would that be placed as it is only really used in `DumpsCommand.java` so `utils/` wouldn't make sense imo.

src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -330,6 +330,8 @@ public String getAuthors() {
         return authors.toString();
     }
 
+
```
benwoo1110, line 333: ahhh whitespace

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -0,0 +1,298 @@
...
+
+        BukkitRunnable logPoster = new BukkitRunnable() {
+            @Override
+            public void run() {
+                HashMap<String, String> pasteURLs = new HashMap<>();
+                Logging.finer("Logs type is: " + logsType);
+                Logging.finer("Services is: " + servicesType);
+
+                // Deal with logs flag
+                if (!parsedFlags.hasFlag("--paranoid")) {
```
benwoo1110, line 150:
```suggestion
                if (!paranoid) {
```

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
benwoo1110: I think separating to methods is fine. Looks to me like you can split "Deal with logs flag" and "Deal with uploading debug info" into separate methods.

benwoo1110 (changes requested):

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
benwoo1110: ohhh right you can do `parsedFlags.flagValue("--upload", ServiceOption.PASTEGG, ServiceOption.class)`

benwoo1110 (approved):

PR #2984, opened 2023-08-29
Worlds config

benwoo1110:
- Move config to new api
- Scaffolding for new world manager class


src/main/java/com/onarandombox/MultiverseCore/MultiverseCore.java
```diff
@@ -124,6 +127,9 @@ public void onEnable() {
             config.setFirstSpawnLocation(firstSpawnWorld.getName());
         }
 
+        var newWorldManager = newWorldManagerProvider.get();
+        newWorldManager.loadAllWorlds();
```
dumptruckman, line 131: can `loadAllWorlds` just be part of the constructor for the new world manager?

benwoo1110: maybe? but i need to ensure the load order first, which is not the priority rn, so I just did the same as before

src/main/java/com/onarandombox/MultiverseCore/worldnew/config/WorldConfigNodes.java
```diff
@@ -0,0 +1,134 @@
...
+    public final ConfigNode<Integer> PLAYER_LIMIT = node(ConfigNode.builder("player-limit", Integer.class)
+            .defaultValue(-1)
+            .name("player-limit")
+            .build());
+
+    public final ConfigNode<AllowedPortalType> PORTAL_FORM = node(ConfigNode.builder("portal-form", AllowedPortalType.class)
+            .defaultValue(AllowedPortalType.ALL)
+            .name("portal-form")
+            .serializer((NodeSerializer<AllowedPortalType>) ENUM_NODE_SERIALIZER)
+            .build());
```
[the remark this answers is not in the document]
benwoo1110: Yea possible, will do

src/test/java/org/mvplugins/multiverse/core/world/WorldConfigTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        compareConfigFile("worlds2.yml", "/delete_worlds.yml")
+    }
+
+    private fun compareConfigFile(configPath: String, comparePath: String) {
+        val config = multiverseCore.dataFolder.toPath().resolve(configPath).toFile().readText()
+        val configCompare = getResourceAsText(comparePath)
+        assertNotNull(configCompare)
+        assertEquals(configCompare, config)
+    }
```
dumptruckman, line 74: When a yaml file is loaded, I don't believe the map keys are in a guaranteed order. If you have more than 1 element in the map, this may fail because of this ordering issue. A way around this is to use something like Hamkrest and `assertThat`.

dumptruckman (dismissed):

dumptruckman (approved):

PR #2986, opened 2023-08-30
Add `/mv coordinates` command

zax71:
Works exactly the same as MV4 with the only change being that the command has been renamed from `/mv coord` to `/mv coordinates` with aliases for the former


src/main/resources/multiverse-core_en.properties
```diff
@@ -19,6 +19,17 @@ mv-core.clone.success=Cloned world '{world}'!
 # /mv confirm
 mv-core.confirm.description=Confirms dangerous commands before executing them.
 
+# /mv coordinates
+mv-core.coordinates.description=Simply sends your coordinates
+mv-core.coordinates.error.players-only=&cThis command can only be used by players
+mv-core.coordinates.error.multiverseWorldOnly=&cThis can only be used in multiverse worlds
```
benwoo1110, line 25: I think this 2 message can be generic error messages like `mv-core.commands.error.mutliverseworldonly`. Bcu seems like it will be used for other commands as well.

src/main/resources/multiverse-core_en.properties
```diff
@@ -19,6 +19,17 @@ mv-core.clone.success=Cloned world '{world}'!
 # /mv confirm
 mv-core.confirm.description=Confirms dangerous commands before executing them.
 
+# /mv coordinates
+mv-core.coordinates.description=Simply sends your coordinates
+mv-core.coordinates.error.players-only=&cThis command can only be used by players
```
benwoo1110, line 24: Having a `-` in the message path may not work, did you test it?

src/main/java/com/onarandombox/MultiverseCore/utils/MVCorei18n.java
```diff
@@ -19,6 +19,16 @@ public enum MVCorei18n implements MessageKeyProvider {
     CLONE_FAILED,
     CLONE_SUCCESS,
 
+    // Coordinates command
+    COORDINATES_ERRORPLAYERSONLY,
```
benwoo1110, line 23:
```suggestion
    COORDINATES_ERROR_PLAYERSONLY,
```

src/main/java/com/onarandombox/MultiverseCore/commands/CoordinatesCommand.java
```diff
@@ -0,0 +1,61 @@
...
+    @Inject
+    public CoordinatesCommand(@NotNull MVCommandManager commandManager, @NotNull MVWorldManager worldManager, @NotNull LocationManipulation locationManipulation) {
+        super(commandManager);
+
+        this.worldManager = worldManager;
+        this.locationManipulation = locationManipulation;
+    }
+
+    @Subcommand("coordinates|coord|coords|co")
+    @CommandPermission("multiverse.core.coordinates")
```
benwoo1110, line 37: I think for now perms node just follow MV4

```suggestion
    @CommandPermission("multiverse.core.coord")
```

src/main/java/com/onarandombox/MultiverseCore/commands/CoordinatesCommand.java
```diff
@@ -0,0 +1,61 @@
...
+    @Description("{@@mv-core.coordinates.description}")
+    public void onCoordinatesCommand(BukkitCommandIssuer issuer) {
+        if (!issuer.isPlayer()) { // Players only
+            issuer.sendInfo(MVCorei18n.COORDINATES_ERRORPLAYERSONLY);
```
benwoo1110, line 41: May want to add support for viewing another players coord with a separate permission node in the future.

src/main/java/com/onarandombox/MultiverseCore/commands/CoordinatesCommand.java
```diff
@@ -0,0 +1,61 @@
...
+            return;
+        }
+
+        Player player = issuer.getPlayer();
```
benwoo1110, line 45: Player should be in the method's parameter. You can use `@Flags("resolve=issuerOnly")`

benwoo1110 (changes requested):

src/main/resources/multiverse-core_en.properties
zax71: Ooo, good point. I’ll swap it over

src/main/resources/multiverse-core_en.properties
zax71: Huh, no it doesn’t work but I’m sure I committed a fix for that 🤔

src/main/java/com/onarandombox/MultiverseCore/commands/CoordinatesCommand.java
zax71: That could be a useful feature, I’ll look in to that.

Is there an example of checking perms or do we just use the Bukkit one?

src/main/resources/multiverse-core_en.properties
benwoo1110: It's not supposed to work, I think so away with the - for now

benwoo1110 (approved):

PR #2989, opened 2023-09-02
Fix /mv dumps to support ANSI and UTF-8 encoded log files

zax71:
Still needs more testing

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
@@ -1,7 +1,6 @@
 package com.onarandombox.MultiverseCore.commands;
 
 import co.aikar.commands.CommandIssuer;
-import co.aikar.commands.InvalidCommandArgument;
 import co.aikar.commands.annotation.CommandAlias;
 import co.aikar.commands.annotation.CommandCompletion;
 import co.aikar.commands.annotation.CommandPermission;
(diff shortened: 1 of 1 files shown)


src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -157,14 +154,27 @@ public void run() {
...
+            return Files.readString(logsPath, StandardCharsets.ISO_8859_1);
         } catch (IOException e) {
-            Logging.warning("Could not read logs/latest.log");
-            throw new RuntimeException(e);
+            // UTF-8 encoded log
+            try {
+                return Files.readString(logsPath, StandardCharsets.UTF_8);
+            } catch (IOException ex) {
+                // It is some other strange encoding
+                throw new RuntimeException(ex);
```
benwoo1110, line 174: Just log servere, printstacetrace and return null

src/main/java/com/onarandombox/MultiverseCore/commands/DumpsCommand.java
```diff
@@ -157,14 +154,27 @@ public void run() {
...
+
 
-        // Try to read file
         try {
-            return Files.readString(logsPath);
+            return Files.readString(logsPath, StandardCharsets.ISO_8859_1);
         } catch (IOException e) {
-            Logging.warning("Could not read logs/latest.log");
-            throw new RuntimeException(e);
+            // UTF-8 encoded log
```
benwoo1110, line 169: Can just be a separate try..catch, don't need to nest them. Bcu if it succeeds it will return already

benwoo1110 (changes requested):

benwoo1110 (approved):

PR #3003, opened 2023-09-09
Loosen up some codestyle

benwoo1110:


config/mv_checks.xml
```diff
@@ -192,7 +192,6 @@
         <module name="PackageDeclaration"/>
         <module name="ParameterAssignment"/>
         <module name="RequireThis"/>
-        <module name="ReturnCount"/>
```
dumptruckman, line 195: Can we just increase the number of permitted returns? Having lots of returns isn't great...

config/mv_checks.xml
```diff
@@ -455,7 +454,7 @@
             <property name="max" value="30"/>
         </module>
         <module name="LambdaBodyLength">
-            <property name="max" value="10"/>
+            <property name="max" value="50"/>
```
dumptruckman, line 457: Dang, 50 is HUGE. I'm okay with increasing this but maybe not quite this much. If a lamda is that big it should probably be broken up with some method calls or something.

dumptruckman (approved):

PR #3008, opened 2023-09-10
Revamp results to support value returns

benwoo1110:


src/main/java/com/onarandombox/MultiverseCore/utils/result/Attempt.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    default Attempt<T, F> peek(Consumer<T> consumer) {
+        if (this instanceof Success) {
+            consumer.accept(get());
+        }
+        return this;
+    }
+
+    default <N> Attempt<N, F> map(Function<T, Attempt<N, F>> mapper) {
```
dumptruckman, line 46: I think this should be `<U> Attempt<U, F> map(Function<? extends T, ? extends U> mapper)`

PR #3019, opened 2023-09-12
Ben/mv5/teleporter

benwoo1110:


src/main/java/org/mvplugins/multiverse/core/api/BlockSafety.java
```diff
@@ -37,43 +38,74 @@ public interface BlockSafety {
...
+
+    /**
+     * Gets the next safe location around the given location.
+     *
+     * @param location  A {@link Location}.
+     * @param tolerance The tolerance.
+     * @param radius    The radius.
+     * @return A safe {@link Location}.
+     */
+    @Nullable Location getSafeLocation(Location location, int tolerance, int radius);
```
dumptruckman, line 64: Could you add some info about what tolerance is in these docs?

src/main/java/org/mvplugins/multiverse/core/commands/DeleteCommand.java
dumptruckman, line 1: How come these commands are in this PR that don't seem related to teleporting? Can we separate into a separate PR maybe?

src/main/java/org/mvplugins/multiverse/core/api/SafeTTeleporter.java
```diff
@@ -14,6 +14,7 @@
 /**
  * Used to safely teleport people.
  */
+@Deprecated
 @Contract
 public interface SafeTTeleporter extends Teleporter {
```
dumptruckman, line 19: Why leave it in at all?

src/main/java/org/mvplugins/multiverse/core/teleportation/TeleportQueue.java
```diff
@@ -12,7 +12,7 @@ public class TeleportQueue {
 
     private final Map<String, String> teleportQueue;
 
-    public TeleportQueue() {
+    TeleportQueue() {
         teleportQueue = new HashMap<>();
```
dumptruckman, line 16: Does it matter that we're not using a thread safe Map implementation here?

src/main/java/org/mvplugins/multiverse/core/teleportation/AsyncSafetyTeleporter.java
```diff
@@ -0,0 +1,164 @@
...
+import org.jetbrains.annotations.Nullable;
+import org.jvnet.hk2.annotations.Service;
+
+import org.mvplugins.multiverse.core.api.BlockSafety;
+import org.mvplugins.multiverse.core.destination.ParsedDestination;
+import org.mvplugins.multiverse.core.utils.result.Result;
+
+@SuppressWarnings("unchecked")
+@Service
+public class AsyncSafetyTeleporter {
```
dumptruckman, line 24: Is it worth distinguishing this as Async if we don't have an alternative?

src/main/java/org/mvplugins/multiverse/core/api/SafeTTeleporter.java
benwoo1110: Bcu of the world manager, now that it’s nuked, it can be removed

src/main/java/org/mvplugins/multiverse/core/teleportation/TeleportQueue.java
benwoo1110: thenAccept for completable future is not on an async thread right?

src/main/java/org/mvplugins/multiverse/core/teleportation/AsyncSafetyTeleporter.java
benwoo1110: Good point

src/main/java/org/mvplugins/multiverse/core/commands/DeleteCommand.java
benwoo1110: The reason why its in this PR, is bcu it was in the old teleporter class, so I cant just remove it since its useful. But i feel it doesnt belong to the teleporter, since just getting a location.

src/main/java/org/mvplugins/multiverse/core/api/BlockSafety.java
benwoo1110: yea sure... this method is just moved, so didnt really check it

src/main/java/org/mvplugins/multiverse/core/commands/DeleteCommand.java
benwoo1110: Oh also.. bcu everything that reference the old teleporter needs to shift

src/test/java/org/mvplugins/multiverse/core/inject/InjectionTest.kt
```diff
@@ -50,12 +48,6 @@ class InjectionTest : TestWithMockBukkit() {
         assertNotNull(multiverseCore.getService(SimpleLocationManipulation::class.java))
     }
 
-    @Test
-    fun `SafeTTeleporter is available as a service`() {
-        assertNotNull(multiverseCore.getService(SafeTTeleporter::class.java))
-        assertNotNull(multiverseCore.getService(SimpleSafeTTeleporter::class.java))
-    }
```
dumptruckman, line 57: Can you add a test to replace this for the new service(s)?

dumptruckman (approved):

PR #3023, opened 2023-09-14
Ben/mv5/modify command

benwoo1110:


src/main/java/org/mvplugins/multiverse/core/commandtools/MVCommandCompletions.java
```diff
@@ -152,4 +160,28 @@ private List<String> getMVWorldNames(BukkitCommandCompletionContext context) {
...
+            return mvWorld.getConfigurablePropertyNames(modifyType);
+        }).getOrElse(Collections.emptyList());
+    }
+
+    private Collection<String> suggestMVWorldPropsValue(BukkitCommandCompletionContext context) {
+        return Try.of(() -> {
+            MultiverseWorld mvWorld = context.getContextValue(MultiverseWorld.class);
+            ConfigModifyType modifyType = context.getContextValue(ConfigModifyType.class);
+            String propertyName = context.getContextValue(String.class);
+            return mvWorld.suggestPropertyValues(modifyType, propertyName, context.getInput());
```
dumptruckman, line 177: Why is a world suggesting property values? This smells like MultiverseWorld is doing too much.

src/main/java/org/mvplugins/multiverse/core/configuration/handle/GenericConfigHandle.java
dumptruckman, line 1: Maybe parts of this class need to be exposed with a new interface with a getter on MultiverseWorld. I think this would help manage SRP a bit more.

dumptruckman (commented):
This is all I can review atm. Gotta lay down.

PR #3027, opened 2023-09-20
Implement autocomplete and string parsing for config handling

benwoo1110:


src/main/java/org/mvplugins/multiverse/core/configuration/functions/NodeSerializer.java
```diff
@@ -0,0 +1,26 @@
+package org.mvplugins.multiverse.core.configuration.functions;
+
+/**
+ * A function that serializes and deserializes objects to and from YAML.
+ *
+ * @param <T>   The type of the object to serialize and deserialize.
+ */
+public interface NodeSerializer<T> {
```
dumptruckman, line 8: I wonder if there's benefit to breaking this into 2 separate functional interfaces for serialization and deserialization...

src/main/java/org/mvplugins/multiverse/core/configuration/node/ValueNode.java
```diff
@@ -28,6 +32,22 @@ public interface ValueNode<T> extends Node {
      */
     @Nullable T getDefaultValue();
 
+    /**
+     * Suggests possible string values for this node.
+     *
+     * @param input The input string.
+     * @return A collection of possible string values.
+     */
+    @NotNull Collection<String> suggest(@Nullable String input);
```
dumptruckman, line 41: It's not clear to me how this method works based on the naming and docs. What does the input have to do with the return value?

src/main/java/org/mvplugins/multiverse/core/config/MVCoreConfig.java
dumptruckman, line 1: Still wondering here if we can maybe create a slimmer interface over ConfigHandle that we return here for dealing with stringy properties instead of mixing those methods with the normal property accessors.

dumptruckman: Maybe something like `ConfigProperties getProperties()` and `ConfigProperties` exposes `setProperty`, `getProperty`, `setPropertyString` and `suggestPropertyValues`. Also, I think `getSuggestedPropertyValues` may be a better name. `suggestPropertyValues` feels very imperative and like it's modifying the class somehow.

src/main/java/org/mvplugins/multiverse/core/configuration/node/ValueNode.java
benwoo1110: aaaa copilot fail

src/main/java/org/mvplugins/multiverse/core/configuration/functions/NodeSerializer.java
benwoo1110: well you cant really have a serialization without a deserialization

dumptruckman: I was just thinking more for the fact that then they could be implemented with lambdas.

PR #3042, opened 2023-11-04
fix help command regex crash

TheJoshue:


benwoo1110:
we are not supporting 1.12 anymore, you can compile a custom build for yourself with the above if you need to fix this issue.

TheJoshue:
> we are not supporting 1.12 anymore, you can compile a custom build for yourself with the above if you need to fix this issue.

Just leave this somewhere for people who are still not aware of this because there are
Especially for those who want their .jar fixed and not to use 20 more plugins to fix every issue of older mc plugin versions

1.8 fixed jar:
https://github.com/TheJoshue/Multiverse-Core/releases/tag/4.1.1

1.12 fixed jar:
https://github.com/TheJoshue/Multiverse-Core/releases/tag/4.2.3

PR #3057, opened 2024-02-24
Add `/mv spawn tp` and `/mv spawn set`

zax71:
Renamed `/mvspawn` to `/mv spawn tp` and `/mvsetspawn` to `/mv spawn set`

### DIFF
--- a/src/main/java/org/mvplugins/multiverse/core/commands/SpawnCommand.java
+++ b/src/main/java/org/mvplugins/multiverse/core/commands/SpawnCommand.java
@@ -14,6 +14,7 @@ import jakarta.inject.Inject;
 import org.bukkit.entity.Player;
 import org.jetbrains.annotations.NotNull;
 import org.jvnet.hk2.annotations.Service;
+import org.mvplugins.multiverse.core.commandtools.MVCommandIssuer;
 import org.mvplugins.multiverse.core.commandtools.MVCommandManager;
 import org.mvplugins.multiverse.core.commandtools.MultiverseCommand;
 import org.mvplugins.multiverse.core.teleportation.AsyncSafetyTeleporter;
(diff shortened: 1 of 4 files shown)


src/main/java/org/mvplugins/multiverse/core/commands/SpawnCommand.java
```diff
@@ -0,0 +1,77 @@
...
+            @Flags("resolve=issuerAware")
+            @Syntax("[player]")
+            @Description("{@@mv-core.spawn.tp.player.description}")
+            Player player
+           ) {
+        // The player is in the world, so it must be loaded
+        LoadedMultiverseWorld world = worldManager.getLoadedWorld(player.getWorld().getName()).getOrNull();
+
+        // Teleport the player
+        safetyTeleporter.teleportSafely(issuer.getIssuer(), player, world.getSpawnLocation());
```
benwoo1110, line 56: need to check if teleport fails

src/main/java/org/mvplugins/multiverse/core/commands/SpawnCommand.java
```diff
@@ -0,0 +1,77 @@
...
+    void onSpawnTpCommand(
+            BukkitCommandIssuer issuer,
+
+            @Flags("resolve=issuerAware")
+            @Syntax("[player]")
+            @Description("{@@mv-core.spawn.tp.player.description}")
+            Player player
+           ) {
+        // The player is in the world, so it must be loaded
+        LoadedMultiverseWorld world = worldManager.getLoadedWorld(player.getWorld().getName()).getOrNull();
```
benwoo1110, line 53: this also, need to check if world is null

benwoo1110 (commented):
i think dont need to change the command format. `/mv spawn` should be teleport to the world's spawn. Then i think `/mv setspawn` is fine? dont need to change to `/mv spawn set`

src/main/java/org/mvplugins/multiverse/core/commands/SpawnCommand.java
zax71: How do I go about doing this? Using `.toAttempt().onFailiure()` causes the server to crash

benwoo1110 (approved):

PR #3077, opened 2024-07-18
Added option to disable saving of source world when cloning.

citylion:
Hello I have never contributed to multiverse before, but I use your plugin as a dependency for my minecraft minigame server. I often clone a template world to open an arena world. Whenever this occurred the server would take a TPS drop. It appears it was from the source world being forcibly saved - though my server was already configured to autosave worlds.

I made a fork to make this saving a configurable operation. If you'd like to incorporate it, here it is.


ccuser44 (approved):

benwoo1110:
This is implemented in a different way for mv5 in #3210

PR #3093, opened 2024-08-27
Allow /mv list to show world names instead of aliases

### LINKED ISSUE #1420: [Suggestion] Include option to switch between worldalias and worldname in /mv list

zax71:
Adds a `--raw` and `-r` flag to allow `/mv list` to show the non-aliased names of worlds

Fixes #1420


zax71:
fixes #1420

zax71:
Argh, it won't let me mark it to auto close that issue

benwoo1110 (approved):

benwoo1110:
I'm just not sure raw is the best wording for this hm

PR #3110, opened 2024-09-21
Internal Chunk Generators and multiverse:void generator as first custom gen. VOID as Environment type when creating or importing worlds. 

MarcoLvr:
Implemented internal Chunk generators. Now multiverse can have its own custom chunk generators. Implemented the void generator (very useful). Added world type "VOID" in environments list. When create or import a world with type VOID the world will be added as a NORMAL world with "multiverse:void" generator.


DaddyRobLIVE:
please review this!

benwoo1110:
There exist many custom generator plugins, including good void generator plugins, my stance on this is unless we can provider a significantly better feature than those plugins, there isn't a need for us to inbuild custom generators into multiverse

PR #3124, opened 2024-11-14
Add latest dir for javadoc

benwoo1110:


.github/workflows/dispatch.javadoc.yml
```diff
@@ -28,12 +28,20 @@ jobs:
...
           GITHUB_VERSION: ${{ inputs.target_tag }}
-      
+
+      - name: Copy javadoc to version and latest folder
+        run: |
+          mkdir -p build/docs/javadoc-deploy/latest
+          mkdir -p build/docs/javadoc-deploy/${{ inputs.target_tag }}
+          cp -r build/docs/javadoc/* build/docs/javadoc-deploy/latest
+          cp -r build/docs/javadoc/* build/docs/javadoc-deploy/${{ inputs.target_tag }}
+
```
dumptruckman, line 38: I don't think we should copy to latest on dispatch. Only the release action should do that.

benwoo1110: hmm, maybe i should add a boolean field?

dumptruckman (approved):

PR #3130, opened 2024-11-20
Refactor to remove need for ParsedDestination and add destination tests

benwoo1110:
ParsedDestination was just like a proxy for DestinationInstance, i realised it was a dumb way of implementing, so i removed it. Need this changes to update multiverse-portals to work with mv5 core.


src/main/java/org/mvplugins/multiverse/core/destination/Destination.java
```diff
@@ -8,7 +8,7 @@
 import org.jvnet.hk2.annotations.Contract;
 
 @Contract
-public interface Destination<T extends DestinationInstance> {
+public interface Destination<D extends Destination<D, T>, T extends DestinationInstance<T, D>> {
```
dumptruckman, line 11: Not crazy about the cursive generics here. Can you summarize what this is doing? Having some difficulty following.

benwoo1110: Its just each Destination maps a one DestinationInstance

benwoo1110: This is so that `ExactDestination.parseDestination()` can only return `ExactDestinationInstance`, then `ExactDestinationInstance.getDestination()` can only return `ExactDestination`. 

Kinda  like a 1 to 1 map thing, where you cant just allow ExactDestination to return any type of DestinationInstance and ExactDestinationInstance be created by any type of Destination

Issue #3153, opened 2025-01-16
[Bug]: `/mv regen -s` command does not update the world's `spawnLocation`


jfoucher:
### `/mv version -p` output

Multiverse-Core Version: 4.3.1-b861

### Server logs

N/A

### Server Version

```shell
paper 1.20.4-499
```

### Bug Description

I have a mining that I regenerate regularly so that the resources are replenished. I also change the seed so that the players can't remember where minerals are. For this purpose I use the command `/mv regen mine -s` which works fine.
The problem is that as soon as I try teleporting in this world, either I get a "no safe location found" message, or I spawn in the middle of water or other unsafe places.
I gather that this is because the `spawnLocation` in the worlds.yml config file remains unchanged after a regen.
I tried editing that file to remove `spawnLocation` but it comes back (although more suited to the new world it appears)
I my mind, regenerating a world with a seed change should also reset the spawnLocation.

### Steps to reproduce

- start a server with multiverse
- create a world `mv create another NORMAL`
- teleport to it `mvtp another`
- you are safe
- `mvtp world`
- `/mv regen another -s`
- `/mv confirm`
- `/mvtp another`
- you might get "no safe location found" or arrive in water or arrive on land.

### Agreements

- [x] I have searched for and ensured there isn't already an open or resolved issue(s) regarding this.
- [x] I was able to reproduce my issue on a freshly setup and up-to-date server with the latest version of Multiverse plugins with no other plugins and with no kinds of other server or client mods.

benwoo1110:
This issue has been fixed in MV5

PR #2824, opened 2023-02-02
MV5

benwoo1110:
# WILL BE UPDATED AS DEVELOPMENT MATERIALISE MORE

## Big Boy Features
- [x] Convert all commands to [ACF](https://github.com/aikar/commands) Command Framework
- [x] Localisation i18n support
- [x] New command flag management api
- [x] Revamp destination api
- [x] Revamp permission checking and handling
- [x] Revamp world manager
- [x] Redo teleporter and teleport queue
- [x] Redo custom MV events being called
- [x] Change to use [CommentedConfiguration](https://github.com/TownyAdvanced/CommentedConfiguration)

## Issues to resolve
- [x] Replace deprecated EntityCreatePortalEvent with PortalCreateEvent (#2000)
- [x]  Loading/Unloading world with command block (#2560)
- [ ] Keep certain world files when regening (#1907, #2783)
- [x] Deprecated API methods (#2436)
- [x] Bypass confirm for dangerous actions (#2764)
- [x] Adds the ability to set fixed player join location (#2461, #2752)
- [ ] Add support for non-default bukkit.yml config location (#2791)
- [x] Format-strings instead of string concatenation (#360)
- [x] Improper usage of groupId and artifactId (#447)
- [x] `/mv anchor <name> <x,y,z>` and `multiverse.core.list.anchor` (#831)
- [ ] Any destination as respawn point (#901)
- [x] Revise how we handle permissions. (#928)
- [ ] Finer permissions for world properties and commands. (#1292)
- [x] List over world properties (#1399)
- [x] Include option to switch between worldalias and worldname in /mv list (#1420)
- [ ] Multiverse Teleport to another world with "~" coords (#1442)
- [ ] Use setSpawnFlags() to control monster spawning (#1469)
- [ ] Per-World Worldborder Command (#1776)
- [x] Add tabcomplete (#1955, #2729)
- [x] Order mv list output alphabetically (#1960)
- [x] Double charge $ on entryfee when teleporting to nether (#1971)
- [x] Better world manager operation result output (#2148)
- [ ] First spawn in a server "ignores" the multiverse set gamemode (#2171)
- [ ] Issue when player relogin to a multiverse world that was deleted (#2221)
- [x] Change bukkit world's spawnpoint to align with mv's (#2226)
- [x] Customise and translate messages (#2273, #2575)
- [ ] Import command switch to automatically extract Nether & End dimensions from a Vanilla world save (#2389)
- [x] Remove messagecooldown and teleportcooldown (#2424)
[...]

PR #3154, opened 2025-01-16
Force spawn location regeneration when regenerating a world with a new seed

jfoucher:
Fixes https://github.com/Multiverse/Multiverse-Core/issues/3153

### DIFF
--- a/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
+++ b/src/main/java/com/onarandombox/MultiverseCore/MVWorld.java
@@ -196,6 +196,15 @@ public class MVWorld implements MultiverseWorld {
         setAllowMonsterSpawn(canMonstersSpawn());
     }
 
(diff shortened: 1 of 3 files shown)


jfoucher:
Doing `world.setSpawnLocation(new MVWorld.NullLocation());` [here](https://github.com/Multiverse/Multiverse-Core/blob/main/src/main/java/com/onarandombox/MultiverseCore/utils/WorldManager.java#L929) should have worked, but the `NullLocation` is transformed to a real location (`Location{world=null,x=0.0,y=64.0,z=0.0,pitch=0.0,yaw=0.0}
`) by the time it arrives to the `MVWorld` constructor, which is why I had to add this public function.

benwoo1110:
hmm i think i fixed this issue with the mv5 alpha branch already

benwoo1110:
Fixed with #2824

PR #3157, opened 2025-01-19
`/mv who` command auto world

zax71:
Sets the world for `/mv who` to the player's world if it is not specified

----
📦 Artifacts generated:
- [multiverse-core-pr3157](https://nightly.link/Multiverse/Multiverse-Core/actions/runs/12871129185/multiverse-core-pr3157.zip)

### DIFF
--- a/src/main/java/org/mvplugins/multiverse/core/commands/WhoCommand.java
+++ b/src/main/java/org/mvplugins/multiverse/core/commands/WhoCommand.java
@@ -69,6 +69,7 @@ public class WhoCommand extends CoreCommand {
         this.worldManager = worldManager;
     }
 
+    @CommandAlias("mvwhoall")
     @Subcommand("whoall")
     @CommandPermission("multiverse.core.list.who.all")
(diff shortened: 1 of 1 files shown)


src/main/java/org/mvplugins/multiverse/core/commands/WhoCommand.java
```diff
@@ -101,6 +103,7 @@ void onWhoAllCommand(
     void onWhoCommand(
             MVCommandIssuer issuer,
 
+            @Optional
             @Syntax("<world>")
             @Description("{@@mv-core.who.world.description}")
             LoadedMultiverseWorld inputtedWorld,
```
benwoo1110, line 109: Can just add a @Flag("resolve=issuerAware") to get the world from player itself

zax71: Oh, I always forget that...

zax71:
That commit is untested

benwoo1110 (approved):

PR #3158, opened 2025-01-23
Ben/mv5/revamp api

benwoo1110:
----
📦 Artifacts generated:
- [multiverse-core-pr3158](https://nightly.link/Multiverse/Multiverse-Core/actions/runs/12932225092/multiverse-core-pr3158.zip)


src/main/java/org/mvplugins/multiverse/core/world/WorldConfigNodes.java
```diff
@@ -118,7 +116,33 @@ private <T> ConfigNode<T> node(ConfigNode.Builder<T, ?> nodeBuilder) {
     final ConfigNode<Biome> BIOME = node(ConfigNode.builder("biome", Biome.class)
             .defaultValue(Biome.CUSTOM)
             .name(null)
-            .serializer(new BiomeSerializer()));
```
dumptruckman, line 121: Why not keep these classes?

benwoo1110: it's one time use, I don't think it needs to be separated

dumptruckman (approved):

PR #3179, opened 2025-03-15
Refactor runQueuedCommand to relay failure reason to caller

benwoo1110:
Use Attempt<> to return failure reason when trying to run a command queue. This also allow use to easily localize the messages.

----
📦 Artifacts generated:
- [multiverse-core-pr3179](https://nightly.link/Multiverse/Multiverse-Core/actions/runs/13868254500/multiverse-core-pr3179.zip)


src/main/java/org/mvplugins/multiverse/core/commandtools/queue/CommandQueueManager.java
```diff
@@ -134,41 +137,43 @@ private Runnable expireRunnable(@NotNull String senderName) {
      * @param issuer    Sender that confirmed the command.
      * @return True if queued command ran successfully, else false.
      */
-    public boolean runQueuedCommand(@NotNull MVCommandIssuer issuer, int otp) {
+    public Attempt<Void, RunQueuedFailedReason> runQueuedCommand(@NotNull MVCommandIssuer issuer, int otpInput) {
```
dumptruckman, line 140: Just now noticing this but `int` doesn't seem like a great type for an otpInput. I would think `String` would make more sense or even maybe more preferably some kind of custom type.

benwoo1110: That's a good point, prob do it in the next pr

PR #3184, opened 2025-03-25
Better handling for a list of attempts with AttemptsAggregate

benwoo1110:
----
📦 Artifacts generated:
- [multiverse-core-pr3184](https://nightly.link/Multiverse/Multiverse-Core/actions/runs/14084819254/multiverse-core-pr3184.zip)


src/main/java/org/mvplugins/multiverse/core/utils/result/AttemptsAggregate.java
```diff
@@ -0,0 +1,64 @@
...
+        return this;
+    }
+
+    public AttemptsAggregate<T, F> onFailure(Consumer<List<Attempt<T, F>>> failureConsumer) {
+        if (failureAttempts.isEmpty()) {
+            return this;
+        }
+        failureConsumer.accept(failureAttempts);
+        return this;
+    }
```
dumptruckman, line 45: These give modifiable access to the underlying list. Consider using Iterable in the Consumer type. Also provide a copy of the list?

benwoo1110: the streams toList() method gives an immutable list, so it will be immutable here already.

dumptruckman (approved):

PR #3266, opened 2025-06-20
Add 1.21.6 as supported version for Modrinth and Hangar

benwoo1110:
----
📦 Artifacts generated:
- [multiverse-core-pr3266](https://nightly.link/Multiverse/Multiverse-Core/actions/runs/15780873466/multiverse-core-pr3266.zip)


zax71:
Does dbo not support 1.21.6 yet?

benwoo1110:
They always take super long to add new mc versions as a valid option

PR #3320, opened 2025-08-17
Fix adventure dep in spigot-based server

MrXiaoM:
From https://github.com/MrXiaoM/SweetMail/issues/51

As you see, Multiverse-Core can not works with my plugin SweetMail in a spigot-based server called `arclight`, which with no `adventure` defaultly.
My plugin SweetMail will adds the four dependencies below into my plugin classloader:
+ `net.kyori:adventure-api:4.22.0`
+ `net.kyori:adventure-platform-bukkit:4.4.0`
+ `net.kyori:adventure-text-serializer-gson:4.22.0`
+ `net.kyori:adventure-text-minimessage:4.22.0`

Multiverse-Core certainly can't find `net/kyori/adventure/text/serializer/plain/PlainTextComponentSerializer` in this situation because my plugin don't need it.

So we should check that is `PlainTextComponentSerializer` exists too.

### DIFF
--- a/src/main/java/org/mvplugins/multiverse/core/utils/text/ChatTextFormatter.java
+++ b/src/main/java/org/mvplugins/multiverse/core/utils/text/ChatTextFormatter.java
@@ -18,7 +18,8 @@ public final class ChatTextFormatter {
     private static final TextFormatter wrapper;
 
     static {
-        if (ReflectHelper.hasClass("net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer")) {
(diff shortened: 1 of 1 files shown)


benwoo1110:
You should configure your plugin to shadow the adventure dependency tho.

MrXiaoM:
> You should configure your plugin to shadow the adventure dependency tho.

This is to be compatible with both Spigot and Paper in just one jar file. And I can't use native adventure platform in Paper if I shadow and relocate the dependencies.

PR #3330, opened 2025-08-29
Configurable server.properties path

ElFrod0:
Some people relocate `server.properties`, just like any other config file, eg. `bukkit.yml`.
It was only configurable for `bukkit.yml` though. This PR fixes it.


benwoo1110 (approved):
Thanks! Today I learnt there is a startup flag to change server.properties location haha.

PR #3373, opened 2025-11-02
Fix spawn location inconsistency

riku6460:
CraftWorld.setSpawnLocation(Location) requires the world in location. https://github.com/PaperMC/Paper/blob/e8c6ba5068287049c890688534963041208c4213/paper-server/src/main/java/org/bukkit/craftbukkit/CraftWorld.java#L338

See #3321


benwoo1110 (approved):
Thanks! Will accept this for now, but I do wonder if we should throw an error instead if the location given is for another world.

PR #3495, opened 2026-07-08
Added Folia Support.

KeyHeld:
Added Folia Support.


benwoo1110:
And you tested that it can support all multiverse features? Such as world creator and all teleport and portal events works?

benwoo1110:
Just adding the Folia scheduler is not enough to make Multiverse-Core fully work on Folia, sorry.

